PR #74, opened 2015-04-17
Rewrite provider addressing design issues

### LINKED ISSUE #50: Tests: Add more coverage for null values
We've had a couple bugs on null values: #36 & #49. The general problem is that we often get the column's type in order to read the value. For example:

1. Get the column type: INTEGER
2. Call sqlite3_column_int64() on a NULL value: converts it to 0

Basically, we need to check for NULL values *before* getting the column type if we're going to read values based on that type.

### LINKED ISSUE #56: Remove unmanaged resources from DbCommand implementation
It is the case that it has become common practice not to dispose of DbCommand objects and there is a lot of code out there that does not dispose DbCommands. No other providers we have looked at have unmanaged resources in DbCommands, which makes this situation not so bad. This is also true for other SQLite provider versions. It is probably better to update this implementation to also not use unmanaged resources in DbCommands.

### LINKED ISSUE #68: Remove DataReader's type inference logic
Instead of mapping the the column type name to CLR types, we'll just give one of the five built-in data types:
* `byte[]`
* `DBNull`
* `double`
* `string`
* `long`

If you ask for a specific type (e.g. via `GetInt32()` or `GetFieldValue<int>()`) we'll do our best to convert the value to that type.

### LINKED ISSUE #72: Use UTF-16 native functions where possible
This will avoid converting every string to and from UTF-8 during P/Invoke improving performance.

bricelam:
Changes:
- Remove unmaged resources from SqliteCommand (Fixes #56)
- Remove type inference logic from SqliteDataReader (Resolves #68)
- Use Unicode versions of native SQLite interfaces (Resolves #72)

Also resolves #50


bricelam:
DNX (both clr and coreclr) crashes hard on the Unicode P/Invoke calls. All the tests pass fine on .NET 4.6 CLR (i.e. no DNX). @davidfowl Any ideas?

bricelam:
Got to the bottom if it. P/Invoke is trying to free memory that's not its to free. :expressionless:

PR #82, opened 2015-05-20
Expose native methods

### LINKED ISSUE #75: Expose handles to native SQLite objects
If we exposed handles to the underlying [sqlite3](http://sqlite.org/c3ref/sqlite3.html) and [sqlite3_stmt](http://sqlite.org/c3ref/stmt.html) objects, users could then work around any native functionality we haven't exposed by doing the native interop themselves. This would also light up interoperability with libraries like [SQLitePCL.raw](https://github.com/ericsink/SQLitePCL.raw).

natemcmaster:
Fix #75

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteCommand.cs
@@ -111,12 +111,12 @@ namespace Microsoft.Data.Sqlite
             {
                 Sqlite3StmtHandle stmt;
                 var rc = NativeMethods.sqlite3_prepare16_v2(
-                    Connection.Handle,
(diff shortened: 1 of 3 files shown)


bricelam:
Sorry, my issue was very vague. This isn't what it meant. Instead, simply expose two new properties:
* SqliteConnection.Handle an IntPtr to _db.DangerousGetHandle()
* SqliteDataReader.Handle an IntPtr to _stmt.DangerousGetHandle()

natemcmaster:
@bricelam thanks for clarifying. I backed out those changes and just added those 2 properties.

bricelam:
:shipit:

Issue #21, opened 2014-04-11
Move to SQLitePCL.raw


bricelam:
To get out of the business of handling the different ways of invoking native code on various platforms, we should consider building on top of ~~[Portable Class Library for SQLite](https://sqlitepcl.codeplex.com) or~~ [ericsink/SQLitePCL.raw](https://github.com/ericsink/SQLitePCL.raw).

mstyura:
@ericsink, @bricelam, @rowanmiller I just want to vote for adding support of WP and Windows 8/10 portable profiles. As I understand from above discussion it is currently not possible to provide it for WP Silverlight execution environment. As I understand so far there are two libs for SQLite which provides ADO.NET like API. The official [System.Data.Sqlite](https://system.data.sqlite.org/index.html/doc/trunk/www/index.wiki) and [Mono.Data.Sqlite](https://github.com/mattleibow/Mono.Data.Sqlite). First one is for full featured .NET and is not suitable for mobile environment unfortunately :( . Second one is intended for mobiles, but it's not obvious how well this library is supported. So it would be nice to have [Microsoft.Data.Sqlite](https://github.com/aspnet/Microsoft.Data.Sqlite/) which is suitable for mobile and universal apps also. Please, consider making it as much portable as possible.

Thanks in advance.

bricelam:
We're primarily limited by the places System.Data or System.Data.Common are available.  Today, that means:
* .NET Framework/Mono
* .NET Core
    * Universal Windows Platform/.NET Native
    * DNX Core/Core CLR
* Xamarin
    * Xamarin.Android
    * Xamarin.iOS
    * Xamarin.iOS (Classic)

mstyura:
@bricelam is the source code of [System.Data.Common](https://www.nuget.org/packages/System.Data.Common/) available in public? BTW, because the lack of `System.Data` in portable profiles, [Mono.Data.SQlite](https://github.com/mattleibow/Mono.Data.Sqlite/tree/master/System.Transactions) even introduces [System.Data.Portable](https://github.com/mattleibow/Mono.Data.Sqlite/tree/master/System.Data) and `System.Transactions.Portable` :|

bricelam:
It's on [the list](https://github.com/dotnet/corefx-progress/blob/master/src-diff/README.md) to be open sourced as part of .NET Core.

ericsink:
[...]

PR #84, opened 2015-05-26
Add shared-cache connection option

natemcmaster:
Add option to enable shared-cache on new connections.
https://www.sqlite.org/sharedcache.html

Also improves the calls added to #82

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnection.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Data;
 using System.Data.Common;
+using System.Runtime.InteropServices;
 using Microsoft.Data.Sqlite.Interop;
 
 namespace Microsoft.Data.Sqlite
@@ -13,7 +14,6 @@ namespace Microsoft.Data.Sqlite
         private const string MainDatabaseName = "main";
 
         private string _connectionString;
-        private SqliteConnectionStringBuilder _connectionStringBuilder;
         private ConnectionState _state;
         private Sqlite3Handle _db;
 
@@ -41,15 +41,17 @@ namespace Microsoft.Data.Sqlite
                 }
 
                 _connectionString = value;
-                _connectionStringBuilder = new SqliteConnectionStringBuilder(value);
+                ConnectionStringBuilder = new SqliteConnectionStringBuilder(value);
             }
         }
+        internal SqliteConnectionStringBuilder ConnectionStringBuilder { get; set; }
 
         public override string Database => MainDatabaseName;
         public override string DataSource =>
             _state == ConnectionState.Open
                 ? NativeMethods.sqlite3_db_filename(_db, MainDatabaseName)
-                : _connectionStringBuilder.DataSource;
+                : ConnectionStringBuilder.DataSource;
+
         public override string ServerVersion => NativeMethods.sqlite3_libversion();
         public override ConnectionState State => _state;
         protected internal SqliteTransaction Transaction { get; set; }
@@ -75,7 +77,10 @@ namespace Microsoft.Data.Sqlite
                 throw new InvalidOperationException(Strings.OpenRequiresSetConnectionString);
             }
 
-            var rc = NativeMethods.sqlite3_open16(_connectionStringBuilder.DataSource, out _db);
+            var rc = NativeMethods.sqlite3_enable_shared_cache(ConnectionStringBuilder.CacheMode == CacheMode.Shared); // called to ensure mode set correctly before a connection is created
+            MarshalEx.ThrowExceptionForRC(rc, _db);
+
+            rc = NativeMethods.sqlite3_open16(ConnectionStringBuilder.DataSource, out _db);
             MarshalEx.ThrowExceptionForRC(rc, _db);
             SetState(ConnectionState.Open);
         }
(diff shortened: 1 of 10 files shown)


bricelam:
Before 442eaab0c38529307767536052bee16b15666740, we were using `sqlite_open_v2()` which allowed us to pass in `SQLITE_OPEN_SHAREDCACHE`. We even had a connection string parameter `Cache=Shared`. Given we want to do #21, we'll probably go back to the non-Unicode overloads anyway...

natemcmaster:
Until #21 is ready, can we add this to improve transactions support in EF? https://github.com/aspnet/EntityFramework/pull/2277

bricelam:
I think we should...
* Remove the property on `SqliteConnection`--the connection string option is enough
* Change the connection string option name to `Cache` with the possible values `Shared` & `Private` to better match the native client

natemcmaster:
@bricelam I've updated this PR with your suggestions. I'll squash to one commit before merging.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -50,10 +53,13 @@ public override string ConnectionString
             _state == ConnectionState.Open
                 ? NativeMethods.sqlite3_db_filename(_db, MainDatabaseName)
                 : _connectionStringBuilder.DataSource;
+
         public override string ServerVersion => NativeMethods.sqlite3_libversion();
         public override ConnectionState State => _state;
         protected internal SqliteTransaction Transaction { get; set; }
 
+        internal bool SharedCacheMode => _sharedCacheMode;
```
bricelam: Remove (unused)

src/Microsoft.Data.Sqlite/Strings.resx
```diff
@@ -138,6 +141,9 @@
   <data name="InvalidIsolationLevel" xml:space="preserve">
     <value>The IsolationLevel '{isolationLevel}' is invalid.</value>
   </data>
+  <data name="InvalidIsolationLevelForUnsharedCache" xml:space="preserve">
+    <value>The IsolationLevel '{isolationLevel}' is invalid when the connection is not using Shared Cache mode.</value>
```
bricelam: Also tell how to fix it. E.g. "Specify 'Cache=Shared' in the connection string."

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionStringBuilderTest.cs
```diff
@@ -172,5 +194,17 @@ public void TryGetValue_returns_true_when_exists()
...
+
+        [Fact]
+        public void ToString_builds_string()
+        {
+            var builder = new SqliteConnectionStringBuilder
+            {
+                DataSource = "test.db",
+                CacheMode = CacheMode.Shared
+            };
+            Assert.Equal("Data Source=test.db;Cache=Shared", builder.ToString());
```
bricelam: Does it serialize if it's set to Private/the default?

src/Microsoft.Data.Sqlite/SqliteConnectionStringBuilder.cs
```diff
@@ -14,27 +14,32 @@ namespace Microsoft.Data.Sqlite
...
-            DataSource
+            DataSource,
+            Cache
         }
 
         private static readonly IReadOnlyList<string> _validKeywords;
         private static readonly IReadOnlyDictionary<string, Keywords> _keywords;
 
         private string _dataSource = string.Empty;
+        private CacheMode _cacheMode = CacheMode.Private;
```
bricelam: There are subtle differences between unspecified and Private:

> Shared cache is disabled by default. But this might change in future releases of SQLite.

Maybe we should add a `Default` option.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -16,6 +17,7 @@ public class SqliteConnection : DbConnection
         private SqliteConnectionStringBuilder _connectionStringBuilder;
         private ConnectionState _state;
         private Sqlite3Handle _db;
+        private bool _sharedCacheMode;
```
bricelam: Technically, this is already cached by `_connectionStringBuilder.CacheMode`, but there's no harm in caching it again.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionStringBuilderTest.cs
natemcmaster: Added test to ensure it does not serialize unset properties.

src/Microsoft.Data.Sqlite/SqliteConnectionStringBuilder.cs
natemcmaster: We will defer adding a `Default` mode until we modify how connections are opened. (switching to open_v2)

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionStringBuilderTest.cs
```diff
@@ -34,13 +42,27 @@ public void DataSource_defaults_to_empty()
         }
 
         [Fact]
+        public void CacheMode_throws_when_invalid_mode()
+        {
+            var ex = Assert.Throws<ArgumentException>(() => new SqliteConnectionStringBuilder("Cache=Valley"));
```
bricelam: :mouse: Cheese?

bricelam:
:shipit:

PR #86, opened 2015-06-02
Workaround for the SQLite download step in our build

natemcmaster:
SQLite download fails on Mono because of https://bugzilla.xamarin.com/show_bug.cgi?id=30686

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@
 bin/
 obj/
 project.lock.json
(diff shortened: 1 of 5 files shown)


natemcmaster:
Should help us get rid of those pesky Travis CI "build failed" emails.

bricelam:
(Transitive :shipit: via #89)

PR #89, opened 2015-06-03
Use sqlite3_open_v2 instead && use UTF-8

### LINKED ISSUE #88: Shared cache does not work on OS X Mono
SqliteConnection.Open() always fails on the OS X default SQLite library and on iPhone 5.

See https://www.sqlite.org/c3ref/enable_shared_cache.html

Source of error:
```c#
NativeMethods.sqlite3_enable_shared_cache(ConnectionStringBuilder.CacheMode == CacheMode.Shared);
```
This always returns SQLITE_MISUSE

natemcmaster:
This allows shared_cache on OS X and iPhone. Also it has the advantage of setting Cache=Shared only for the currently opened connection and does not tamper with global settings.

Fixes #88.

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
@@ -90,8 +90,24 @@ namespace Microsoft.Data.Sqlite.Interop
             }
         }
 
-        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode, ExactSpelling = true)]
-        public static extern int sqlite3_bind_text16(Sqlite3StmtHandle pStmt, int i, string zData, int n, IntPtr xDel);
+        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
+        private static extern int sqlite3_bind_text(Sqlite3StmtHandle pStmt, int i, IntPtr zData, int n, IntPtr xDel);
+
+        public static int sqlite3_bind_text(Sqlite3StmtHandle pStmt, int i, string data, int n, IntPtr xDel)
+        {
+            var zData = MarshalEx.StringToHGlobalUTF8(data);
+            try
+            {
+                return sqlite3_bind_text(pStmt, i, zData, n, xDel);
+            }
+            finally
+            {
+                if (zData != IntPtr.Zero)
+                {
+                    Marshal.FreeHGlobal(zData);
+                }
+            }
+        }
 
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         public static extern int sqlite3_changes(Sqlite3Handle db);
@@ -124,10 +140,10 @@ namespace Microsoft.Data.Sqlite.Interop
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         public static extern int sqlite3_column_count(Sqlite3StmtHandle stmt);
 
-        [DllImport("sqlite3", EntryPoint = "sqlite3_column_decltype16", CallingConvention = CallingConvention.Cdecl)]
-        public static extern IntPtr sqlite3_column_decltype16_raw(Sqlite3StmtHandle stmt, int iCol);
+        [DllImport("sqlite3", EntryPoint = "sqlite3_column_decltype", CallingConvention = CallingConvention.Cdecl)]
+        public static extern IntPtr sqlite3_column_decltype_raw(Sqlite3StmtHandle stmt, int iCol);
 
-        public static string sqlite3_column_decltype16(Sqlite3StmtHandle stmt, int iCol) => Marshal.PtrToStringUni(sqlite3_column_decltype16_raw(stmt, iCol));
+        public static string sqlite3_column_decltype(Sqlite3StmtHandle stmt, int iCol) => MarshalEx.PtrToStringUTF8(sqlite3_column_decltype_raw(stmt, iCol));
 
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         public static extern double sqlite3_column_double(Sqlite3StmtHandle stmt, int iCol);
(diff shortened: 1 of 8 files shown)


bricelam:
How does this affect the other `sqlite_*16` calls? Should they be changed too?

natemcmaster:
Good point. Unfortunately yes it will affect them. I'll update those too. From SQLite.org:
>The default encoding will be UTF-8 for databases created using sqlite3_open() or sqlite3_open_v2(). The default encoding for databases created using sqlite3_open16() will be UTF-16 in the native byte order.

natemcmaster:
Converted all calls to UTF-8 since we can't use open16

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -90,8 +90,11 @@ public static int sqlite3_bind_parameter_index(Sqlite3StmtHandle pStmt, string z
             }
         }
 
-        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode, ExactSpelling = true)]
-        public static extern int sqlite3_bind_text16(Sqlite3StmtHandle pStmt, int i, string zData, int n, IntPtr xDel);
+        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
+        private static extern int sqlite3_bind_text(Sqlite3StmtHandle pStmt, int i, IntPtr zData, int n, IntPtr xDel);
+
+        public static int sqlite3_bind_text(Sqlite3StmtHandle pStmt, int i, string data, int n, IntPtr xDel) =>
+            sqlite3_bind_text(pStmt, i, MarshalEx.StringToHGlobalUTF8(data), n, xDel);
```
bricelam: We need to free this memory.

bricelam: See [the previous implementation](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/ff35414ba1d88340bbdc195c2306a97b10d52a6d/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs#L70-L88).

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -188,22 +188,29 @@ public static string sqlite3_db_filename(Sqlite3Handle db, string zDbName)
...
 
-        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode, ExactSpelling = true)]
-        public static extern int sqlite3_open16(string filename, out Sqlite3Handle ppDb);
+        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
+        private static extern int sqlite3_open_v2(IntPtr filename, out Sqlite3Handle ppDb, int flags, IntPtr vfs);
+
+        public static int sqlite3_open_v2(string filename, out Sqlite3Handle ppDb, int flags, string vfs)
+        {
+            var zFilename = MarshalEx.StringToHGlobalUTF8(filename);
+            var zVfs = string.IsNullOrEmpty(vfs) ? IntPtr.Zero : MarshalEx.StringToHGlobalUTF8(vfs);
```
bricelam: Free these too.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -111,7 +111,7 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
             do
             {
                 Sqlite3StmtHandle stmt;
-                var rc = NativeMethods.sqlite3_prepare16_v2(
+                var rc = NativeMethods.sqlite3_prepare_v2(
                     Connection.DbHandle,
                     tail,
                     -1,
```
bricelam: May get a small perf gain by passing in the index of the null terminator.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -78,10 +78,10 @@ public override void Open()
                 throw new InvalidOperationException(Strings.OpenRequiresSetConnectionString);
             }
 
-            var rc = NativeMethods.sqlite3_enable_shared_cache(ConnectionStringBuilder.CacheMode == CacheMode.Shared); // called to ensure mode set correctly before a connection is created
-            MarshalEx.ThrowExceptionForRC(rc, _db);
+            var flags = Constants.SQLITE_OPEN_READWRITE | Constants.SQLITE_OPEN_CREATE;
```
bricelam: Isn't read-write-create the default anyway?

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -78,10 +78,10 @@ public override void Open()
...
+            flags |= (ConnectionStringBuilder.CacheMode == CacheMode.Shared) ? Constants.SQLITE_OPEN_SHAREDCACHE : Constants.SQLITE_OPEN_PRIVATECACHE;
 
-            rc = NativeMethods.sqlite3_open16(ConnectionStringBuilder.DataSource, out _db);
+            var rc = NativeMethods.sqlite3_open_v2(ConnectionStringBuilder.DataSource, out _db, flags, null);
```
bricelam: For literals, we tend to specify the parameter name to make it more readable. e.g `vfs: null`

bricelam:
:shipit:

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: It's the default behavior of sqlite3_open and sqlite3_open16, but https://sqlite.org/c3ref/open.html doesn't seem clear to me about what happens if I pass in 0 for the flags parameter. I figure it's safer to be explicit in defining the behavior we want.

bricelam: Heh, it is pretty vague... Setting it looks like the right thing to do.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -111,10 +112,10 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
             do
             {
                 Sqlite3StmtHandle stmt;
-                var rc = NativeMethods.sqlite3_prepare16_v2(
+                var rc = NativeMethods.sqlite3_prepare_v2(
                     Connection.DbHandle,
                     tail,
-                    -1,
+                    Encoding.UTF8.GetByteCount(tail) + 1, // add one to include the nul-terminator. See SQLite.org docs for sqlite3_prepare_v2.
```
bricelam: I'm not sure re-calculating this makes it worth it. :smile: The previous code set it lazily just after converting it to UTF8.

bricelam: More context:
The only perf it saves in the native code is scanning the string for the null terminator. Converting the string to UTF8 + getting the would-be UTF8 byte count is probably more expensive.

natemcmaster: Yeah, I noticed that within sqlite3_prepare_v2 already calls Encoding.UTF8.GetBytes. We could save the native call a little work by using the result of this.

bricelam: :+1:

bricelam: P.S. We're undoubtedly over-optimizing here, but I've done that obsessively on this entire project , so why stop now? :wink:.

natemcmaster: ...haha. I was just thinking--is it faster for our managed code to find and use bytes.Length than for native code to find the null-terminator? I'm not sure. Maybe we need some real perf tests, not just our intuition.

natemcmaster:
Rebased and added perf fix to sqlite_prepare_v2

PR #91, opened 2015-06-05
Add Travis-CI status images to README.md

natemcmaster:
Only reason I can think not to put this on here: it only tests Mono.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 Microsoft.Data.Sqlite
 =====================
 
+AppVeyor: [![AppVeyor build status](https://ci.appveyor.com/api/projects/status/p48patmrpydigrj0?svg=true)](https://ci.appveyor.com/project/aspnetci/microsoft-data-sqlite)
+
+Travis: [![Travis build Status](https://travis-ci.org/aspnet/Microsoft.Data.Sqlite.svg?branch=dev)](https://travis-ci.org/aspnet/Microsoft.Data.Sqlite)
+
 Contains SQLite implementations of the System.Data.Common interfaces.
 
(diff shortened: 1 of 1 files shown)


bricelam:
We should be consistent with the other repos. It look like [MVC's README](https://github.com/aspnet/Mvc/blob/dev/README.md) just points to the dev branch. We should also have an AppVeyor status--I think it adds non-Mono coverage.

natemcmaster:
Definitely agree on being consistent, but I would like to make the readme clearer on which branch the CI label references, especially because we use 'dev' as our primary branch.

Also, are we using AppVeyor? I would have expected the project to be at https://ci.appveyor.com/project/aspnetci/entityframework/branch/dev, but it's not there.

NTaylorMullen:
We debated showcasing different branches and came up with the following: The release branch doesn't really make sense to showcase. It only ever revs once a release (therefore only ever fails briefly) so adding it would be unneeded noise.

As for consistency, we should definitely be staying consistent with MVC/other repos. Heck we even had a table at first showcasing appveyor/GitHub and then reverted it :smile:.

Lastly to get things setup on appveyor talk to @Eilon and he can set everything up on that end to get the process flowing.

Eilon:
Yeah I have a mail thread to enable AppVeyor for this repo. Will try to get that done today.

As far as the labels/badges go: yeah, `dev` only.

NTaylorMullen:
Looks good. Will need to add AppVeyor once Eilon enables it.

Eilon:
AppVeyor should now be enabled for this repo.

PR #92, opened 2015-06-10
Update CacheMode.cs

Eilon:
Ignore this


Eilon:
@natemcmaster @bricelam ok it seems to show the two Travis builds and the AppVeyor build here.

PR #98, opened 2015-06-18
Fix memory leak.

natemcmaster:
When a SQLite statement causes an error sqlite3_step, the statement object was not disposed.

Add test cases to replicate https://github.com/aspnet/EntityFramework/issues/2356.

Improved error messages for SQL errors

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/Constants.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/Constants.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Data.Sqlite.Interop
     internal static class Constants
     {
         public const int SQLITE_OK = 0;
-        
+
         public const int SQLITE_ROW = 100;
         public const int SQLITE_DONE = 101;
 
@@ -17,12 +17,11 @@ namespace Microsoft.Data.Sqlite.Interop
         public const int SQLITE_TEXT = 3;
         public const int SQLITE_BLOB = 4;
         public const int SQLITE_NULL = 5;
-        
+
         public const int SQLITE_OPEN_SHAREDCACHE = 0x00020000;
         public const int SQLITE_OPEN_PRIVATECACHE = 0x00040000;
         public const int SQLITE_OPEN_READWRITE = 0x00000002;
         public const int SQLITE_OPEN_CREATE = 0x00000004;
-        
 
         public static readonly IntPtr SQLITE_TRANSIENT = new IntPtr(-1);
     }
--- a/src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Data.Sqlite.Interop
                 ? NativeMethods.sqlite3_errstr(rc)
                 : NativeMethods.sqlite3_errmsg(db);
 
-            throw new SqliteException("Error: " + message, rc);
+            throw new SqliteException(Strings.FormatSqliteNativeError(rc, message), rc);
         }
     }
 }
--- a/src/Microsoft.Data.Sqlite/SqliteDataReader.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteDataReader.cs
@@ -9,6 +9,7 @@ using System.Diagnostics;
 using System.Globalization;
 using Microsoft.Data.Sqlite.Interop;
 using Microsoft.Data.Sqlite.Utilities;
+using static Microsoft.Data.Sqlite.Interop.Constants;
 #if NET45 || DNX451
 using System.Data;
 
@@ -96,7 +97,7 @@ namespace Microsoft.Data.Sqlite
             var rc = NativeMethods.sqlite3_step(_stmt);
             MarshalEx.ThrowExceptionForRC(rc, _db);
 
-            _done = rc == Constants.SQLITE_DONE;
+            _done = rc == SQLITE_DONE;
 
             return !_done;
         }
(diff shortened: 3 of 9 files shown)


bricelam:
Wouldn't the GC just free it anyway when it goes out of scope?

natemcmaster:
Apparently not. This was the reason scratch *.db files could not be deleted at the end of EF tests.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
```diff
@@ -94,9 +95,17 @@ public override bool Read()
             }
 
             var rc = NativeMethods.sqlite3_step(_stmt);
-            MarshalEx.ThrowExceptionForRC(rc, _db);
+            try
+            {
+                MarshalEx.ThrowExceptionForRC(rc, _db);
+            }
+            catch
```
bricelam: Is this one needed? If callers put the `DbDataReader` is in a `using` block (like they always should) throwing would cause this (and its statement handle) to be immediately disposed.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
```diff
@@ -9,6 +9,7 @@
 using System.Globalization;
 using Microsoft.Data.Sqlite.Interop;
 using Microsoft.Data.Sqlite.Utilities;
+using static Microsoft.Data.Sqlite.Interop.Constants;
```
bricelam: Hmm, we should discuss this one as a team.

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
```diff
@@ -64,7 +64,7 @@ public static void ThrowExceptionForRC(int rc, Sqlite3Handle db)
                 ? NativeMethods.sqlite3_errstr(rc)
                 : NativeMethods.sqlite3_errmsg(db);
 
-            throw new SqliteException("Error: " + message, rc);
+            throw new SqliteException(Strings.FormatSqliteNativeError(rc, message), rc);
```
bricelam: My reasoning for this was that it's the format returned by sqlite3.exe. I didn't bother with a resource since SQLite itself isn't localized.

test/Microsoft.Data.Sqlite.Tests/SqliteConcurrencyTest.cs
```diff
@@ -0,0 +1,96 @@
...
+        public SqliteConcurrencyTest()
+        {
+            using (var connection = CreateConnection())
+            {
+                connection.Open();
+                var command = connection.CreateCommand();
+                command.CommandText = @"CREATE TABLE IF NOT EXISTS a (b);
+										INSERT INTO a VALUES (1);
+										INSERT INTO a VALUES (2);
+										";
```
bricelam: TABS! :fearful:

src/Microsoft.Data.Sqlite/Interop/Constants.cs
```diff
@@ -17,12 +17,16 @@ internal static class Constants
...
         public const int SQLITE_OPEN_SHAREDCACHE = 0x00020000;
         public const int SQLITE_OPEN_PRIVATECACHE = 0x00040000;
         public const int SQLITE_OPEN_READWRITE = 0x00000002;
         public const int SQLITE_OPEN_CREATE = 0x00000004;
-        
+
+        // Error codes
+        public const int SQLITE_ERROR = 1;
+        public const int SQLITE_BUSY = 5;
+        public const int SQLITE_LOCKED = 6;
```
bricelam: Should these be moved to the test project? They're never used by product code.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
natemcmaster: I added it in case it isn't in a `using` block, even though they should.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
natemcmaster: IMO, this is good for classes of constants. Otherwise, it obscures code.

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
natemcmaster: Yeah, I just added it to be consistent with how we generate exception messages.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
bricelam: I think we can rely on the contract of the class for this--that consumers must dispose. Their app has bigger problems if they're leaving readers open.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
bricelam: *I'm* fine with it. @Eilon do we (aspnet) have any guidance on this new language feature yet?

bricelam:
:shipit:

natemcmaster:
Updated to respond to feedback.
cc @Eilon I can revert the `using static` if that is against aspnet style guides.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
Eilon: No general guidance on this one. I think it really depends on how clear the code is. I think for example in most apps `using static System.Console` would *not* be very clear because `WriteLine` is not very meaningful on its own. But also, *meh* :smile:

PR #100, opened 2015-07-07
Support UWP (and .NET Native)

bricelam:
This uses the NuGet 3 runtimes directory to get the native library into the app. This will require a NuGet 3 client to install the package.

There are currently some bugs around CLR AnyCPU projects that prevent the package from being installed. I'll work with the NuGet team on these.


makefile.shade
```diff
@@ -50,18 +37,30 @@ var sqliteDllPath='src/Microsoft.Data.Sqlite/redist/x86/sqlite3.dll'
             {
                 CreatePartFromFile(
                     package,
-                    @"src/Microsoft.Data.Sqlite/build/net45/Microsoft.Data.Sqlite.targets",
-                    @"build/net45/Microsoft.Data.Sqlite.targets");
+                    sqliteDllPath,
+                    @"runtimes/win-x86/native/sqlite3.dll");
+                CreatePartFromFile(
+                    package,
+                    "src/Microsoft.Data.Sqlite/runtimes/win/native/x86/sqlite3.dll",
```
natemcmaster: Doesn't match with sqliteDllpath. Should it?

bricelam: It could, but doesn't need to. Per lines 21 & 22, they're both copies of the same thing.

natemcmaster:
:shipit:

PR #103, opened 2015-07-17
Implement command timeouts 

### LINKED ISSUE #97: Handle SQLITE_BUSY and SQLITE_LOCKED
We do not currently handle the error code SQLITE_LOCKED (https://www.sqlite.org/rescode.html#locked). We could use the sqlite3_unlock_notify API to create blocking calls to sqlite3_step and sqlite3_prepare_v2. This should prevent the error from occurring. https://www.sqlite.org/unlock_notify.html

natemcmaster:
Implements a timeout option on the connection string. Default timeout = 30s

Users can modify this with the connection string `Data Source =test.db; Timeout=10 `

This handles SQLITE_LOCKED and SQLITE_BUSY, and throws a `TimeoutException` if they are not handled within the timeout specified in the connection string.

Works for both cache modes (Shared and Private).

TODO: perf

Fixes #97

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
@@ -4,6 +4,9 @@
 using System;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
+using System.Threading;
+using static Microsoft.Data.Sqlite.Interop.Constants;
+
 #if NET45 || DNX451 || DNXCORE50
 using Microsoft.Framework.Internal;
 
@@ -68,6 +71,12 @@ namespace Microsoft.Data.Sqlite.Interop
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         private static extern int sqlite3_bind_parameter_index(Sqlite3StmtHandle pStmt, IntPtr zName);
 
+        [DllImport("sqlite3", EntryPoint = "sqlite3_bind_parameter_name", CallingConvention = CallingConvention.Cdecl)]
+        private static extern IntPtr sqlite3_bind_parameter_name_raw(Sqlite3StmtHandle stmt, int i);
+
+        public static string sqlite3_bind_parameter_name(Sqlite3StmtHandle stmt, int i)
+            => MarshalEx.PtrToStringUTF8(sqlite3_bind_parameter_name_raw(stmt, i));
+
         public static int sqlite3_bind_parameter_index(Sqlite3StmtHandle pStmt, string zName)
         {
             var ptr = MarshalEx.StringToHGlobalUTF8(zName);
@@ -102,6 +111,9 @@ namespace Microsoft.Data.Sqlite.Interop
                 }
             }
         }
+
+        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
+        public static extern int sqlite3_busy_timeout(Sqlite3Handle db, int ms);
 
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         public static extern int sqlite3_changes(Sqlite3Handle db);
@@ -243,15 +255,32 @@ namespace Microsoft.Data.Sqlite.Interop
         }
 
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
+        public static extern int sqlite3_reset(Sqlite3StmtHandle stmt);
+
+        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         public static extern int sqlite3_step(Sqlite3StmtHandle stmt);
+
+        //TODO move to a utility class
+        public static int sqlite3_step_blocking(Sqlite3Handle db, Sqlite3StmtHandle stmt, int milliseconds)
+        {
+            var timer = new Stopwatch();
+            int rc;
+            timer.Start();
+            while (SQLITE_LOCKED == (rc = sqlite3_step(stmt)))
+            {
+                if (timer.ElapsedMilliseconds >= milliseconds)
+                {
+                    return rc;
+                }
+                sqlite3_reset(stmt);
+                Thread.Sleep(150);
+                // TODO add async code path that uses Task.Delay() instead
+            }
+            return rc;
+        }
 
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         public static extern int sqlite3_stmt_readonly(Sqlite3StmtHandle pStmt);
 
-        [DllImport("sqlite3", EntryPoint = "sqlite3_bind_parameter_name", CallingConvention = CallingConvention.Cdecl)]
-        private static extern IntPtr sqlite3_bind_parameter_name_raw(Sqlite3StmtHandle stmt, int i);
-
-        public static string sqlite3_bind_parameter_name(Sqlite3StmtHandle stmt, int i)
-            => MarshalEx.PtrToStringUTF8(sqlite3_bind_parameter_name_raw(stmt, i));
     }
 }
--- a/src/Microsoft.Data.Sqlite/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteCommand.cs
@@ -5,10 +5,12 @@ using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Data.Sqlite.Interop;
+using static Microsoft.Data.Sqlite.Interop.Constants;
 
 namespace Microsoft.Data.Sqlite
 {
@@ -69,7 +71,7 @@ namespace Microsoft.Data.Sqlite
 
         public new virtual SqliteParameterCollection Parameters => _parameters.Value;
         protected override DbParameterCollection DbParameterCollection => Parameters;
-        public override int CommandTimeout { get; set; }
+        public override int CommandTimeout { get; set; } = 30;
         public override bool DesignTimeVisible { get; set; }
         public override UpdateRowSource UpdatedRowSource { get; set; }
         public new virtual SqliteParameter CreateParameter() => new SqliteParameter();
@@ -104,6 +106,9 @@ namespace Microsoft.Data.Sqlite
                         : Strings.TransactionConnectionMismatch);
             }
 
+            //TODO not necessary to call every time a command is executed. Only on first command or when timeout changes
+            NativeMethods.sqlite3_busy_timeout(Connection.DbHandle, CommandTimeout * 1000);
+
             var hasChanges = false;
             var changes = 0;
             var stmts = new Queue<Tuple<Sqlite3StmtHandle, bool>>();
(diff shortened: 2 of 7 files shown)


src/Microsoft.Data.Sqlite/Interop/NativeMethods`.cs
```diff
@@ -0,0 +1,200 @@
...
+using Microsoft.Framework.Runtime;
+using Microsoft.Framework.Runtime.Infrastructure;
+using Microsoft.Framework.DependencyInjection;
+#endif
+
+#endif
+
+namespace Microsoft.Data.Sqlite.Interop
+{
+    internal static partial class NativeMethods
```
bricelam: Why split this into two files?

src/Microsoft.Data.Sqlite/Interop/NativeMethods`.cs
```diff
@@ -0,0 +1,200 @@
...
+            timer.Start();
+            while (SQLITE_LOCKED == (rc = sqlite3_step(stmt)))
+            {
+                if (timer.ElapsedMilliseconds >= milliseconds)
+                {
+                    return rc;
+                }
+                sqlite3_reset(stmt);
+                Thread.Sleep(150);
+            }
```
bricelam: Is the retry loop still needed?

src/Microsoft.Data.Sqlite/Interop/NativeMethods`.cs
```diff
@@ -0,0 +1,200 @@
...
+            finally
+            {
+                Marshal.FreeHGlobal(zSqlPtr);
+            }
+        }
+
+        public static int sqlite3_prepare_v2_blocking(Sqlite3Handle db, string zSql, out Sqlite3StmtHandle ppStmt, out string pzTail, int milliseconds)
+        {
+            sqlite3_busy_timeout(db, milliseconds);
+            return sqlite3_prepare_v2(db, zSql, out ppStmt, out pzTail);
```
bricelam: I'd prefer to keep NativeMethods 1:1 with the native APIs. I.e. Move this to SqliteCommand.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -111,7 +113,13 @@ protected override void Dispose(bool disposing)
         }
 
         // NB: Other providers don't set Transaction
-        public new virtual SqliteCommand CreateCommand() => new SqliteCommand { Connection = this, Transaction = Transaction };
+        public new virtual SqliteCommand CreateCommand() => new SqliteCommand
+        {
+            Connection = this,
+            Transaction = Transaction,
+            CommandTimeout = ConnectionStringBuilder.Timeout
```
bricelam: I don't think we should have a connection string parameter for this.

src/Microsoft.Data.Sqlite/SqliteConnectionStringBuilder.cs
```diff
@@ -15,11 +15,13 @@ public class SqliteConnectionStringBuilder : DbConnectionStringBuilder
     {
         private const string DataSourceKeyword = "Data Source";
         private const string CacheKeyword = "Cache";
+        private const string TimeoutKeyword = "Timeout";
```
bricelam: We shouldn't have a connection string setting for this. It changes over the lifetime of the connection, and there is already API for setting it: DbCommand.CommandTimeout.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -111,11 +114,16 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
...
-                    Connection.DbHandle,
-                    tail,
-                    out stmt,
-                    out tail);
+                var rc = NativeMethods.sqlite3_prepare_v2_blocking(
+                        Connection.DbHandle,
+                        tail,
+                        out stmt,
+                        out tail,
+                        CommandTimeout*1000);
```
bricelam: This is cleaner than I thought it would be. Only one command ever runs on the connection at a time, so the fact that the timeout is actually associated with the connection doesn't ever really mater.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -30,6 +32,7 @@ public SqliteCommand(string commandText, SqliteConnection connection)
             : this(commandText)
         {
             Connection = connection;
+            CommandTimeout = connection.ConnectionStringBuilder.Timeout;
```
bricelam: Should default to 30.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -111,11 +114,16 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
...
-                    out tail);
+                var rc = NativeMethods.sqlite3_prepare_v2_blocking(
+                        Connection.DbHandle,
+                        tail,
+                        out stmt,
+                        out tail,
+                        CommandTimeout*1000);
+                if (rc == SQLITE_BUSY)
+                {
+                    throw new TimeoutException(Strings.FormatCommandTimeout(CommandTimeout));
```
bricelam: Don't throw a `TimeoutException`; anything other than a `SqliteException` would be unexpected to callers. Just let the next line throw.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -154,9 +162,14 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
                     throw new InvalidOperationException(Strings.FormatMissingParameters(string.Join(", ", unboundParams)));
                 }
 
-                rc = NativeMethods.sqlite3_step(stmt);
                 try
                 {
+                    rc = NativeMethods.sqlite3_step_blocking(Connection.DbHandle, stmt, CommandTimeout*1000);
```
bricelam: This is cleaner than I thought it would be. Only one command ever runs on the connection at a time, so the fact that the timeout is actually associated with the connection doesn't ever really mater.

src/Microsoft.Data.Sqlite/Interop/NativeMethods`.cs
natemcmaster: The busy handler only works for SQLITE_BUSY. SQLITE_LOCKED on the other hand can only be handled with unlock_notify or this retry loop.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
natemcmaster: True. We could just call the sqlite3_busy_timeout as soon as the connection is opened and leave it that way for all native calls.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -111,11 +114,16 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
             do
             {
                 Sqlite3StmtHandle stmt;
-                var rc = NativeMethods.sqlite3_prepare_v2(
-                    Connection.DbHandle,
-                    tail,
-                    out stmt,
-                    out tail);
+                var rc = NativeMethods.sqlite3_prepare_v2_blocking(
```
bricelam: Does `prepare_v2` ever return busy/locked? That seems strange to me...

natemcmaster:
I can backout the connstr option and make this only configurable on the command. Honoring this timeout means we will need to change the busy_timeout for each command execution.

src/Microsoft.Data.Sqlite/Interop/NativeMethods`.cs
natemcmaster: I split it into one file with pure extern declarations and the other for those that perform some marshaling. The file was beginning to feel unreadable.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
```diff
@@ -94,7 +98,11 @@ public override bool Read()
                 return _hasRows;
             }
 
-            var rc = NativeMethods.sqlite3_step(_stmt);
+            var rc = NativeMethods.sqlite3_step_blocking(Command.Connection.DbHandle, _stmt, Command.CommandTimeout*1000);
```
bricelam: Can this return busy/locked? Or just on the first step in SqliteCommand?

natemcmaster: Good point. I hadn't tested with just the first call being blocking. I'll test for it.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
natemcmaster: Yes, it will occasionally return busy. I couldn't find a case of it returning locked.

natemcmaster:
Updated.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -104,6 +106,8 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
                         : Strings.TransactionConnectionMismatch);
             }
 
+            NativeMethods.sqlite3_busy_timeout(Connection.DbHandle, CommandTimeout * 1000);
```
bricelam: Maybe add a TODO to only do this if it's changed since the last time we called it on this open connection.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -243,15 +255,30 @@ public static int sqlite3_prepare_v2(Sqlite3Handle db, string zSql, out Sqlite3S
         }
 
         [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
+        public static extern int sqlite3_reset(Sqlite3StmtHandle stmt);
+
+        [DllImport("sqlite3", CallingConvention = CallingConvention.Cdecl)]
         public static extern int sqlite3_step(Sqlite3StmtHandle stmt);
 
+        public static int sqlite3_step_blocking(Sqlite3Handle db, Sqlite3StmtHandle stmt, int milliseconds)
```
bricelam: It does feel a little weird to have implementation here that doesn't deal with native interop. We may want to move this closer to the places that use it. Or just in a different utility class. But it's also fine to just leave it here for now.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -243,15 +255,30 @@ public static int sqlite3_prepare_v2(Sqlite3Handle db, string zSql, out Sqlite3S
...
+            int rc;
+            timer.Start();
+            while (SQLITE_LOCKED == (rc = sqlite3_step(stmt)))
+            {
+                if (timer.ElapsedMilliseconds >= milliseconds)
+                {
+                    return rc;
+                }
+                sqlite3_reset(stmt);
+                Thread.Sleep(150);
```
bricelam: Can of worms: We could now have an async code path that calls `Task.Delay()` here instead. If nothing else, we should create an issue.

bricelam:
:shipit: Well done on thoroughly investigating and fixing this issue! Many developers will now graciously take it for granded. :wink:

PR #108, opened 2015-07-29
Remove dependency on System.Threading.Thread

### LINKED ISSUE #107: System.Threading.Thread not available on UWP
Trying to install the latest package on UWP yields the following error.

> System.Threading.Thread 4.0.0-beta-23019 provides a compile-time reference assembly for System.Threading.Thread on UAP,Version=v10.0, but there is no run-time 
assembly compatible with win10

natemcmaster:
Fixes #107


bricelam:
:shipit:

PR #109, opened 2015-08-03
Add runtime directive for .NET Native

natemcmaster:
Adds metadata for marshalling handles through interop.

### DIFF
--- a/src/Microsoft.Data.Sqlite/Properties/Microsoft.Data.Sqlite.rd.xml
+++ b/src/Microsoft.Data.Sqlite/Properties/Microsoft.Data.Sqlite.rd.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
+  <Library>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Sqlite/Properties/Microsoft.Data.Sqlite.rd.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
+  <Library>
+    <Type Name="Microsoft.Data.Sqlite.Interop.Sqlite3Handle" MarshalObject="Required All" />
+    <Type Name="Microsoft.Data.Sqlite.Interop.Sqlite3StmtHandle" MarshalObject="Required All" />
```
bricelam: What exactly does this tell it to do? How is it different from the default behavior?

natemcmaster:
I did further testing. This seems to be unnecessary. The errors I experienced were not due to missing metadata. Closing this PR.

PR #113, opened 2015-08-08
Skip NativeLibraryLoader on non-Windows platforms

### LINKED ISSUE #101: Doesn't load native library on OS X Core CLR
`dnvm install -u latest -r coreclr`

On CoreCLR for OS X, NativeLibraryLoader cannot import "api-ms-win-core-libraryloader-l1-1-0".

```
DllNotFoundException: Unable to load DLL 'api-ms-win-core-libraryloader-l1-1-0': The specified module could not be found.
```

natemcmaster:
Fix #101

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
@@ -26,6 +26,12 @@ namespace Microsoft.Data.Sqlite.Interop
 #if NET45 || DNX451 || DNXCORE50
         static NativeMethods()
         {
+#if DNXCORE50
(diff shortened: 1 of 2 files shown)


bricelam:
Can you replace [the code in NativeLibraryLoader](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs#L97-L100) instead?

bricelam:
Or does it fail before that because it tries to JIT the P/Invoke methods?

natemcmaster:
It fails because of https://github.com/aspnet/dnx/issues/2442

bricelam:
Hmm, given that the DNX codepaths here are just temporary anyway... :shipit:

PR #127, opened 2015-09-01
Fix bug in parameter binding with UTF8 strings

natemcmaster:
Fixes https://github.com/aspnet/EntityFramework/issues/2936


bricelam:
:shipit: Well done

PR #129, opened 2015-09-01
Update packages

bricelam:


natemcmaster:
:shipit:

bricelam:
DNX isn't handling this well. Will reconsider later.

PR #136, opened 2015-09-10
Add DbProviderFactory implementation

### LINKED ISSUE #115: Implement DbProviderFactory
Is there any plan to add DbProviderFactory support to this package and/or vnext in general? I am attempting to write a DNX based app that needs interchangeable SQL Server and Sqlite support, and I am using Dapper for data access. Is there any general guidance that you can offer for this type of scenario?

natemcmaster:
Fix #115

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteFactory.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteFactory.cs
@@ -1,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Data.Common;
+
+namespace Microsoft.Data.Sqlite
+{
+    public class SqliteFactory : DbProviderFactory
+    {
+        static SqliteFactory()
+        {
+            Instance = new SqliteFactory();
+        }
+
+        private SqliteFactory()
+        {
+        }
+
+        public readonly static SqliteFactory Instance;
+
+        public override DbCommand CreateCommand()
+            => new SqliteCommand();
+
+        public override DbConnection CreateConnection()
+            => new SqliteConnection();
+
+        public override DbConnectionStringBuilder CreateConnectionStringBuilder()
+            => new SqliteConnectionStringBuilder();
+
+        public override DbParameter CreateParameter()
+            => new SqliteParameter();
+    }
+}
--- a/test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
+++ b/test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
@@ -13,11 +13,11 @@ namespace Microsoft.Data.Sqlite
         [Fact]
         public void Ctor_sets_connection_string()
         {
-            var connectionSring = "Data Source=test.db";
+            var connectionString = "Data Source=test.db";
 
-            var connection = new SqliteConnection(connectionSring);
+            var connection = new SqliteConnection(connectionString);
 
-            Assert.Equal(connectionSring, connection.ConnectionString);
+            Assert.Equal(connectionString, connection.ConnectionString);
         }
 
         [Fact]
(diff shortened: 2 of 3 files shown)


src/Microsoft.Data.Sqlite/SqliteProviderFactory.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Data.Common;
+
+namespace Microsoft.Data.Sqlite
+{
+    public class SqliteProviderFactory : DbProviderFactory
```
bricelam: Just `SqliteFactory`.

bricelam: Make this a singleton with a private constructor and a static `Instance` property.

natemcmaster: Done.

src/Microsoft.Data.Sqlite/SqliteFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public static SqliteFactory Instance
+        {
+            get
+            {
+                if (_instance == null)
+                {
+                    _instance = new SqliteFactory();
+                }
+                return _instance;
```
bricelam: Instead of making the property lazy, just set it in a static constructor. The code will be simpler, and initialization will still happen lazily the first time the type is used (i.e. the property is accessed)

bricelam:
Please add new tests instead of retrofitting into existing ones. Tests should be clear about the thing they are testing, and they should, ideally, only test one thing.

natemcmaster:
updated

src/Microsoft.Data.Sqlite/SqliteFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+        static SqliteFactory()
+        {
+            Instance = new SqliteFactory();
+        }
+
+        private SqliteFactory()
+        {
+        }
+
+        public static SqliteFactory Instance { get; }
```
bricelam: Oh yuck, I just noticed [the other ones](https://msdn.microsoft.com/en-us/library/vstudio/system.data.sqlclient.sqlclientfactory.instance) are ***fields***. We should probably conform in case something in .Full NET depends on that.

For a bit more context, you can register these things in `Web.config`...
```XML
<system.data>
  <DbProviderFactories>
    <add invariant="Bubblegum"
         type="Microsoft.Data.Sqlite.SqliteFactory" />
  </DbProviderFactories>
</system.data>
```
...then load them from code.
```C#
var factory = DbProviderFactories.GetFactory("Bubblegum");
```

natemcmaster: Ah, didn't know this magic was possible. I'll update.

divega: It is old school magic :smile: BTW, I do agree things in .NET might by convention find the `Instance` field using reflection so if we are going to add it is better to follow the same pattern. @ajcvickers might have other insights.

src/Microsoft.Data.Sqlite/SqliteFactory.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Data.Common;
+
+namespace Microsoft.Data.Sqlite
+{
+    public class SqliteFactory : DbProviderFactory
```
divega: FWIW, many providers use "nickname" + Factory to name this class, e.g. `SqlClientFactory`, `OracleClientFactory` (both the obsolete Microsoft and the Oracle implementation use this), `MySqlClientFactory`, `NpgsqlFactory`. If we care to have shorter handle than Microsoft.Data.Sqlite that people can use in StackOverflow.com and other site, this might be a good way to start :smile:

src/Microsoft.Data.Sqlite/SqliteFactory.cs
ajcvickers: The old EF stack looks for the Instance field. At some point I'm pretty sure I updated it to look for a property if the field wasn't found. But if we want to be as compatible as possible with this convention, then we should use a field.

natemcmaster:
Updated to use a field instead of property.

bricelam:
:shipit:

PR #137, opened 2015-09-10
Run tests against coreclr on darwin/linux

### LINKED ISSUE #124: Run tests against CoreCLR on (not) Windows
Presently our build script skips tests on CoreCLR for {darwin, linux}. We should enable this when DNX for these platforms stabilizes.

natemcmaster:
Fix #124


bricelam:
:shipit: You should file an issue on [Universe](https://github.com/aspnet/Universe) for this too.

PR #140, opened 2015-09-17
Minor API tweaks. Add "Filename" synonym to connection string builder.

natemcmaster:
Add "Filename" alias to connection string builder
Rename CacheMode to SqliteConnectionCacheMode.

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConnectionStringBuilder.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnectionStringBuilder.cs
@@ -17,7 +17,9 @@ namespace Microsoft.Data.Sqlite
     public class SqliteConnectionStringBuilder : DbConnectionStringBuilder
     {
         private const string DataSourceKeyword = "Data Source";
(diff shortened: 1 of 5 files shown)


divega:
Looks good to me but will leave it for @bricelam to signoff.

src/Microsoft.Data.Sqlite/SqliteConnectionStringBuilder.cs
```diff
@@ -18,6 +18,7 @@ public class SqliteConnectionStringBuilder : DbConnectionStringBuilder
     {
         private const string DataSourceKeyword = "Data Source";
         private const string CacheKeyword = "Cache";
+        private const string FilenameKeyword = "Filename";
```
bricelam: I think we also wanted *DataSource* to work.

bricelam:
:shipit:

src/Microsoft.Data.Sqlite/SqliteConnectionStringBuilder.cs
natemcmaster: Will add.

PR #141, opened 2015-09-17
Ignore CommandBehavior.SequentialAccess

### LINKED ISSUE #114: ExecuteReader throws exception when invoked with command behavior other than Default (CommandBehavior.Default)
The present implementation of SQLiteCommand and SQLiteDataReader does not support any command behavior but "Default":

The code throws an exception thus causing issues integrating w/ Dapper.NET:
https://github.com/aspnet/Microsoft.Data.Sqlite/blob/e29261c97c1c89567c6bee2211d4fbdc19921828/src/Microsoft.Data.Sqlite/SqliteCommand.cs#L88

natemcmaster:
Fix #114

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteCommand.cs
@@ -88,19 +88,22 @@ namespace Microsoft.Data.Sqlite
 
         public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
         {
-            if (behavior != CommandBehavior.Default)
+            if ((behavior & ~(CommandBehavior.Default | CommandBehavior.SequentialAccess)) != 0)
             {
                 throw new ArgumentException(Strings.FormatInvalidCommandBehavior(behavior));
             }
+            
             if (Connection == null
                 || Connection.State != ConnectionState.Open)
             {
                 throw new InvalidOperationException(Strings.FormatCallRequiresOpenConnection("ExecuteReader"));
             }
+            
             if (string.IsNullOrEmpty(CommandText))
             {
                 throw new InvalidOperationException(Strings.FormatCallRequiresSetCommandText("ExecuteReader"));
             }
+            
             if (Transaction != Connection.Transaction)
             {
                 throw new InvalidOperationException(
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -88,19 +88,22 @@ public override void Prepare()
 
         public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
         {
-            if (behavior != CommandBehavior.Default)
+            if (behavior != CommandBehavior.Default && behavior != CommandBehavior.SequentialAccess)
             {
                 throw new ArgumentException(Strings.FormatInvalidCommandBehavior(behavior));
```
divega: Nit: maybe this should be NotSupportedException, but no big deal.

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
```diff
@@ -297,6 +301,7 @@ public override bool IsDBNull(int ordinal)
 
         public override double GetDouble(int ordinal)
         {
+            AssertSequentialAccess(ordinal);
```
divega: Is it right to assume that the only effect of using `SequentialAccess` is that we force the _`GetXyzz()`_ methods to throw if you read out of order? 

The value of `SequentialAccess` is that it can help remove unnecessary overhead in certain provider implementations. I would recommend just ignoring the behavior (as I think @bricelam originally suggested) rather than adding unnecessary overhead :smile:

bricelam: +1 for simply ignoring the flag.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -88,19 +88,22 @@ public override void Prepare()
 
         public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
         {
-            if (behavior != CommandBehavior.Default)
+            if (behavior != CommandBehavior.Default && behavior != CommandBehavior.SequentialAccess)
```
bricelam: This is a flags enum. The logic here works by coincidence since `Default` is `0`, but there may be a clearer way to express it. Something like...
```C#
behavior & ~(CommandBehavior.Default | CommandBehavior.SequentialAccess) != 0
```
```

natemcmaster:
Updates: i've stripped this down to just ignoring SequentialAccess.

bricelam:
:shipit:

PR #150, opened 2015-10-12
React to aspnet/Universe#290 fix

dougbu:
- get latest `build.cmd` and `build.sh` files
- go back to Mono Beta feed (version 4.0.4) in Travis builds
  - avoid frequent `mono .nuget/nuget.exe` failures
  - skip tests that fail with this Mono version
- remove `test-coreclr` target
  - had incorrect case for `coreclr` runtime and 186 tests fail when using DNX Core

### DIFF
--- a/test/Microsoft.Data.Sqlite.Tests/SqliteConnectionStringBuilderTest.cs
+++ b/test/Microsoft.Data.Sqlite.Tests/SqliteConnectionStringBuilderTest.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.AspNet.Testing.xunit;
 using Xunit;
 
 namespace Microsoft.Data.Sqlite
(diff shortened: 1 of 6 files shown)


natemcmaster:
:shipit:

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionStringBuilderTest.cs
```diff
@@ -174,13 +175,19 @@ public void ShouldSerialize_returns_false_when_not_exists()
             Assert.False(new SqliteConnectionStringBuilder().ShouldSerialize("Invalid"));
         }
 
-        [Fact]
+        [ConditionalFact]
+        [FrameworkSkipCondition(
+            RuntimeFrameworks.Mono,
```
natemcmaster: As a future improvement to testing, it would be better to skip this only on Mono 4.0.2 to 4.0.4, not all of mono.

PR #123, opened 2015-08-31
Detect OS using DNX

natemcmaster:
Workaround aspnet/dnx#2594.
Fixes #121 

The code is right and should work, but cannot verify on OS X until DNX fixes other major issues with osx/coreclr.

PR #156, opened 2015-10-21
React to changes in dnx runtime services

natemcmaster:


anpete:
:shipit:

davidfowl:
Instead use the clr type RuntimeInformation

natemcmaster:
@davidfowl  Do stable packages (`4.0.0`, no star) exist for System.Runtime.InteropServices.RuntimeInformation? We had to work around this in the past https://github.com/aspnet/Microsoft.Data.Sqlite/pull/123

natemcmaster:
Checked...RuntimeInformation will not work for us due to packaging issues. Sticking with this implementation. It will be removed anyways once NuGet supports native binaries.

PR #159, opened 2015-10-23
[testing] Fail all tests if sqlite3.dll cannot be imported

natemcmaster:
Add debugging info to display the version of SQLite used.

Currently, if sqlite3.dll cannot import, then we get 239 errors, all with the same message. This short-circuits and cancels all testing if sqlite3.dll does not import.


divega:
:shipit:

Issue #55, opened 2014-08-31
Use better location as base for relative paths


sgtfrankieboy:
When trying to load a SQLite database from a relative path (`Content\Db\Assets.db`) it will try to load the file from the current process location which is klr.exe instead of the location of the working directory.

**Workarounds**

* Use an absolute path
* Use `PathResolver.ResolveAppRelativePath` from the Configuration project.

natemcmaster:
I can't repo this bug. Was it fixed with #74 ?

natemcmaster:
@sgtfrankieboy are you still experiencing this bug?

bricelam:
In general, we still need to solve the problem. The experience is particularly bad on UWP apps where the default relative path is not writable. We should make things relative to the DataDirectory domain property too, if it's set. Also, if I change the current directory, it shouldn't affect the relative path used in the connection string.

natemcmaster:
Cannot test UWP support until System.Data.Common works for netcore50

natemcmaster:
Instead of doing a magic substitution in the path, I'm in favor of throwing better error messages. On UWP, we throw `Sqlite Error 14: could not open database file`, which doesn't help explain that the folder is inaccessible. We could translate this into a UnauthorizedAccessException which would help users [searching](https://www.bing.com/search?q=windows+store+app+create+file+unauthorizedaccessexception&FORM=AWRE) for a [solution](https://msdn.microsoft.com/en-us/library/windows/apps/hh967755.aspx).

Also, we could provide samples that demonstrate correctly accessing writable folders. E.g.
```c#
new SqliteConnection($"Data Source={ApplicationData.Current.LocalFolder.Path}/test.db");
```

natemcmaster:
@bricelam Can we close this then as 'won't fix'?

bricelam:
@rowanmiller? I'm on the fence. Having relative paths broken by default on UWP still feels like a pit of failure we could address.

rowanmiller:
@natemcmaster @bricelam let's discuss with the team in triage

rowanmiller:
DNX issues are all resolved now. We may consider doing something better for UWP in the future if we see folks hitting issues.

natemcmaster:
*Originally made by @leak on https://github.com/aspnet/EntityFramework/issues/2864*

=================================

Related to https://github.com/aspnet/EntityFramework/issues/2748

Specifying a relative path to a database file does not work, the application dies with 
[...]

PR #161, opened 2015-10-26
Automatically adjust for filesystem defaults on UWP

### LINKED ISSUE #116: Transaction fails due to invalid temporary directory on UWP
(Not sure whether this library is meant to be usable in UAP, but based on https://github.com/aspnet/dnx/issues/2302 I suppose that yes. Currently it works fine when referenced directly).

In UAP you need to adjust SQLite temp path (see similar issue https://github.com/oysteinkrog/SQLite.Net-PCL/pull/157). Without it, certain scenarios involving DDL inside a transaction fail with "SQL logic error".

Related test case:

    public MainPage() {
        this.InitializeComponent();

        var dbPath = Path.Combine(Windows.Storage.ApplicationData.Current.LocalFolder.Path, Guid.NewGuid().ToString());

        // >>>>> this line is a fix
        //SetDirectory(2, Windows.Storage.ApplicationData.Current.TemporaryFolder.Path);

        using(var conn = new SqliteConnection("data source=" + dbPath)) {
            conn.Open();
            using(var cmd = conn.CreateCommand()) {
                cmd.CommandText = @"
                    create table foo(id integer primary key, value integer);
                    insert into foo(value) values ('a');

                    begin; -- >>>>> this causes 'sql logic error'

                    create table foo_copy(id integer primary key, value text);
                    insert into foo_copy select * from foo;
                ";
                cmd.ExecuteNonQuery();
            }
        }
    }

    [DllImport("sqlite3", EntryPoint = "sqlite3_win32_set_directory", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode)]
    static extern int SetDirectory(uint directoryType, string directoryPath);

natemcmaster:
Changes
 - Set temporary folder on connection open on UWP
 - Adjust relative paths to local folder by default

Adds test project to solution.

Fix #116.
Fixes relative path issue on UWP (see #55)


natemcmaster:
cc @bricelam

bricelam:
Does it work with .NET Native?

natemcmaster:
@bricelam Yes, it works.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -90,10 +95,66 @@ public override void Open()
...
+            }
+        }
+
+        private static dynamic GetApplicationData()
+        {
+            var appDataType = Type.GetType("Windows.Storage.ApplicationData, Windows, ContentType=WindowsRuntime", throwOnError: false);
+            var appData = (dynamic)appDataType?.GetTypeInfo()?
+                .GetDeclaredProperty("Current")?
+                .GetMethod?
+                .Invoke(null, null);
```
bricelam: All this null propagation seems excessive. Just return null if `GetType` returns null. The rest should never be null.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -90,10 +95,66 @@ public override void Open()
...
+
+        private string AdjustForRelativeDirectory(string path)
+            => path;
+#else
+        private string AdjustForRelativeDirectory(string path)
+        {
+            var appData = GetApplicationData();
+            try
+            {
+                if (appData == null || appData?.LocalFolder?.Path == null || Path.IsPathRooted(path))
```
bricelam: The middle condition shouldn't be needed--if appData isn't null, the rest won't be null (unless the user jumps through a lot of hoops)

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -90,10 +95,66 @@ public override void Open()
...
+                {
+                    return path;
+                }
+
+                return Path.GetFullPath(Path.Combine(appData.LocalFolder.Path, path));
+            }
+            catch (NotSupportedException)
```
bricelam: What throws this?

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -90,10 +95,66 @@ public override void Open()
...
+            if (appData == null)
+            {
+                return;
+            }
+
+            using (var command = CreateCommand())
+            {
+                command.CommandText = $"PRAGMA temp_store_directory = '{appData.TemporaryFolder.Path}'";
+                command.ExecuteNonQuery();
+            }
```
bricelam: I think there's an existing extension method that does all this

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -90,10 +95,66 @@ public override void Open()
...
+            var appData = GetApplicationData();
+
+            if (appData == null)
+            {
+                return;
+            }
+
+            using (var command = CreateCommand())
+            {
+                command.CommandText = $"PRAGMA temp_store_directory = '{appData.TemporaryFolder.Path}'";
```
bricelam: I'm not a fan of string interpolation--it just adds overhead if you never actually format. I'd use concatenation.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -90,10 +95,66 @@ public override void Open()
...
+                if (appData == null || appData?.LocalFolder?.Path == null || Path.IsPathRooted(path))
+                {
+                    return path;
+                }
+
+                return Path.GetFullPath(Path.Combine(appData.LocalFolder.Path, path));
+            }
+            catch (NotSupportedException)
+            {
+                Debug.WriteLine("Could not adjust relative path for use on UWP.");
```
bricelam: `Debug.Fail`?

uwp/Microsoft.Data.Sqlite.UWP/Microsoft.Data.Sqlite.UWP.csproj
```diff
@@ -0,0 +1,188 @@
...
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{D1E1F777-AF5F-4EF2-9100-6156D4D5FF1D}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Microsoft.Data.Sqlite</RootNamespace>
+    <AssemblyName>Microsoft.Data.Sqlite</AssemblyName>
```
bricelam: Why is this project needed? Can't we reference the DNX output?

uwp/Microsoft.Data.Sqlite.FunctionalTests.UWP/.gitignore
```diff
@@ -0,0 +1,4 @@
+bin/
+obj/
+
+*.pfx
```
bricelam: We should just add this to the root `.gitignore` file

uwp/Microsoft.Data.Sqlite.UWP/Microsoft.Data.Sqlite.UWP.csproj
davidfowl: :+1:

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: GetFullPath. If a user tried to specify a valid [URI filename](https://www.sqlite.org/c3ref/open.html#urifilenameexamples), e.g. `file://( localfolder )/data.db`

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: No need to fail. There are some cases where this will fail and the user-specified path will work.

uwp/Microsoft.Data.Sqlite.UWP/Microsoft.Data.Sqlite.UWP.csproj
natemcmaster: Maybe I'm doing something wrong, but I could not get the UWP test project to reference the DNX class library. I had to add this as a Universal class library for the UWP testing.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: The compiler cannot do extension methods from within the class they extend.

natemcmaster:
Updated with PR feedback.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
bricelam: You have to qualify with `this` or not use the extension method syntax `MyExtensions.Method(this, arg)`

natemcmaster: Ok. Updated.

bricelam:
:shipit:

Issue #134, opened 2015-09-04
Unable to load DLL 'sqlite3' when using packages.config


bricelam:
Because NuGet only copies runtime artifacts when using `project.json`, if you try to use the package on a project using `packages.config`, you'll get the following error.

> Unable to load DLL 'sqlite3': The specified module could not be found. (Exception from HRESULT: 0x8007007E)

We might be able to make this work using MSBuild scripts...

PR #162, opened 2015-10-28
Remove NativeLibraryLoader

### LINKED ISSUE #128: Remove NativeLibraryLoader paths on DNX
~~Blocked on aspnet/dnx#402~~
~~Blocked on aspnet/dnx#3119~~

natemcmaster:
No longer necessary as NuGet and DNX should handle native libraries in the 'runtimes' folder.

Verified on UWP. It will load from "runtimes/win7-(arch)".

Fixes #128 

Note it does not solve #134


src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -6,42 +6,11 @@
...
-#endif
 
 namespace Microsoft.Data.Sqlite.Interop
 {
     internal static class NativeMethods
     {
-#if NET451 || DOTNET5_4
-        static NativeMethods()
-        {
-            var loaded = NativeLibraryLoader.TryLoad("sqlite3");
```
bricelam: Wait, we still need this on Full .NET! Only UWP and DNX will handle the runtimes stuff.

natemcmaster:
@bricelam Is there a difference between the native library we pull from https://www.sqlite.org/2015/sqlite-uap-3090000.vsix vs https://www.sqlite.org/2015/sqlite-dll-win32-x86-3090000.zip?

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
natemcmaster: Oh, I was under the impression nuget 3 would handle this. Does it not automatically bring in native libraries?

bricelam:
Yes, the vsix is compiled for WinRT

natemcmaster:
Jumped the :gun:. Will leave this to @bricelam as fixing #128 is fairly nuanced.

PR #163, opened 2015-10-28
Copy sqlite3.dll when using packages.config

### LINKED ISSUE #134: Unable to load DLL 'sqlite3' when using packages.config

bricelam:
Resolves #134


davidfowl:
Tricky, won't this collide with the default msbuild logic to copy native assets?

bricelam:
It only kicks in if `project.json` is present. This only runs if it's not. :wink:

davidfowl:
ah hah!

bricelam:
...and only on net451 projects.

natemcmaster:
:shirt: :it:

bricelam:
Recalling. I had to rework this a bit for #128.

PR #168, opened 2015-11-04
Automatically adjust relative paths

### LINKED ISSUE #55: Use better location as base for relative paths

natemcmaster:
Always adjust relative paths to the ApplicationBasePath.

Fix #55

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnection.cs
@@ -4,11 +4,12 @@
 using System;
 using System.Data;
 using System.Data.Common;
-#if NETCORE50
-using System.Diagnostics;
 using System.IO;
+#if NETCORE50
 using System.Reflection;
 using Microsoft.Data.Sqlite.Utilities;
+#else
+using Microsoft.Extensions.PlatformAbstractions;
 #endif
 using Microsoft.Data.Sqlite.Interop;
 
@@ -17,7 +18,7 @@ namespace Microsoft.Data.Sqlite
     /// <summary>
     /// Represents a connection with a SQLite database.
     /// </summary>
-    public class SqliteConnection : DbConnection
+    public partial class SqliteConnection : DbConnection
     {
         private const string MainDatabaseName = "main";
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -96,61 +98,88 @@ public override void Open()
...
+            {
+                path = AdjustForRelativeDirectory(path);
+            }
+            
             var rc = NativeMethods.sqlite3_open_v2(path, out _db, flags, vfs: null);
             MarshalEx.ThrowExceptionForRC(rc, _db);
             SetState(ConnectionState.Open);
-
+            
             SetFolders();
```
bricelam: A `partial` method could be used here since it's a no-op outside of netcore50.

natemcmaster:
Rebased and updated.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -105,52 +111,76 @@ public override void Open()
...
-            var appData = (dynamic)appDataType?.GetTypeInfo()
-                .GetDeclaredProperty("Current").GetMethod.Invoke(null, null);
-            return appData;
+            try
+            {
+                return Path.GetFullPath(Path.Combine(DefaultBasePath, path));
+            }
+            catch (Exception ex)
+            {
+                if (ex is ArgumentException || ex is NotSupportedException)
```
bricelam: Could use C# 6 `catch...when` syntax.

bricelam: Or `where` or `if`... whatever it ended up being called :smile:

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -105,52 +111,76 @@ public override void Open()
...
+                return path;
+            }
 
-            if (appData == null)
+            if (path.StartsWith("file:"))
             {
-                return;
+#if NETCORE50
+                // UWP cannot adjust for URI on UWP
+                return path;
```
bricelam: If it's absolute, will it work? If so, we could support this by parsing it using the `Uri` class and fixing up the path.

bricelam:
:shipit:

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: I tried this at first. Even with an absolute path, Uri file paths do not work on UWP.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: C#6 ftw.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
bricelam: I think you'd need to call `sqlite3_config(SQLITE_CONFIG_URI, 1)` first.

bricelam: But it seems pretty low priority...

natemcmaster: The only a reason a user would need Uri syntax is for [query parameters](http://www.sqlite.org/uri.html#coreqp) we don't support. We could add this to the connection string build itself if needed, but for now I agree--this is low priority.

bricelam: Named, shared-cache, in-memory databases are the only scenarios I find interesting here. :smile_cat:

PR #170, opened 2015-11-04
Fix native library loading

### LINKED ISSUE #128: Remove NativeLibraryLoader paths on DNX

### LINKED ISSUE #134: Unable to load DLL 'sqlite3' when using packages.config

bricelam:
Also includes cleanup from the dnx451 to net451 change and other minor refactorings.

Resolves #128, fixes #134

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
@@ -4,40 +4,21 @@
 using System;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
-using static Microsoft.Data.Sqlite.Interop.Constants;
 
-#if NET451 || DOTNET5_4
-using System.IO;
-using System.Reflection;
-using Microsoft.Extensions.PlatformAbstractions;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Internal;
+#if NET451
+using Microsoft.Data.Sqlite.Utilities;
 #endif
 
 namespace Microsoft.Data.Sqlite.Interop
 {
     internal static class NativeMethods
     {
-#if NET451 || DOTNET5_4
+#if NET451
         static NativeMethods()
         {
+            // NOTE: This is used by AnyCPU applications to load the native library based on the process's
+            //       architecture. It does nothing on DNX and Mono which handle native library loading themselves.
             var loaded = NativeLibraryLoader.TryLoad("sqlite3");
-
-            // TODO: Remove when DNX supports native artifacts
-            if (!loaded)
-            {
-                var library = PlatformServices.Default.LibraryManager
-                    .GetLibrary(typeof(NativeMethods).GetTypeInfo().Assembly.GetName().Name);
-
-                var installPath = library.Path;
-                if (library.Type == "Project")
-                {
-                    installPath = Path.GetDirectoryName(installPath);
-                }
-
-                loaded = NativeLibraryLoader.TryLoad("sqlite3", Path.Combine(installPath, "runtimes/win/native"));
-            }
-
             Debug.Assert(loaded, "loaded is false.");
         }
 #endif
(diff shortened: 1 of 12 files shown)


src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -4,40 +4,19 @@
...
 
 namespace Microsoft.Data.Sqlite.Interop
 {
     internal static class NativeMethods
     {
-#if NET451 || DOTNET5_4
+#if NET451
         static NativeMethods()
         {
             var loaded = NativeLibraryLoader.TryLoad("sqlite3");
```
natemcmaster: For which scenarios do we still need this?

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
```diff
@@ -103,6 +80,8 @@ public static bool IsLoaded(string dllName)
             return handle != IntPtr.Zero;
         }
 
+        private static bool IsDNX() => PlatformServices.Default != null;
```
natemcmaster: Won't this always be true with its [current implementation](https://github.com/aspnet/dnx/blob/0b4140fd6827acc912eed858431922574cd3f9f2/src/Microsoft.Extensions.PlatformAbstractions/PlatformServices.cs#L9-L17)?

bricelam: Hmm, it seems to work, but I'll change it to `PlatformServices.Default?.Application.RuntimeFramework.Identifier == "DNX"` just to be safe.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
bricelam: Full .NET (non-DNX)

bricelam: AnyCPU doesn't know its architecture until runtime.

natemcmaster: Ok, that’s what I was missing. Can we add into the comments?The MSBuild fixes this for full .NET when platform arch is known, right?

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
bricelam: Ah, it changed in RC2... That's why it's currently working for me :smile:

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
```diff
@@ -128,11 +125,6 @@ public override bool NextResult()
 #if NET451
         // TODO: Remove when fixed in System.Data.Common
         public override void Close() => Dispose(true);
-
-        public override DataTable GetSchemaTable()
```
natemcmaster: Unfortunately we still need all these NET451 implementations b/c we build against Mono for testing.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
bricelam: It currently *always* copies both the x86 and x64 version and loads the correct one at runtime. We could optimize it to copy the architecture-specific one if it's know at build time, but I didn't bother...

src/Microsoft.Data.Sqlite/SqliteDataReader.cs
bricelam: Ah, is that what the failures are? I thought they updated to net451 contracts...

natemcmaster:
:shipit: (assuming you have approval)

PR #171, opened 2015-11-05
Support SQLite 3.7.9

natemcmaster:
SQLite 3.7.9 is the default library version on Travis Linux agents.
3.7.13 is the default version on aspnet-docker (Fix https://github.com/aspnet/aspnet-docker/issues/85) and Travis OSX agents.

Fallback to older sqlite3 calls when using older versions of sqlite3.

sqlite3_close and sqlite3_errstr do not exist < 3.7.15
sqlite3_db_filename does not exist < 3.7.10

Add internal utilities for working with different sqlite versions and testing.
Add OSX and Mono alpha to test matrix.

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
@@ -60,11 +60,7 @@ namespace Microsoft.Data.Sqlite.Interop
                 return;
             }
 
-            var message = db == null || db.IsInvalid
-                ? NativeMethods.sqlite3_errstr(rc)
-                : NativeMethods.sqlite3_errmsg(db);
-
-            throw new SqliteException(Strings.FormatSqliteNativeError(rc, message), rc);
+            throw new SqliteException(VersionedMethods.SqliteErrorMessage(rc, db), rc);
         }
     }
 }
--- a/src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
@@ -1,0 +1,74 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Data.Sqlite.Interop
+{
+    internal class VersionedMethods
+    {
+        public static string SqliteErrorMessage(int rc, Sqlite3Handle db)
+        {
+            var message = db == null || db.IsInvalid
+                ? _strategy.ErrorString(rc)
+                : NativeMethods.sqlite3_errmsg(db);
+
+            return Strings.FormatSqliteNativeError(rc, message);
+        }
+
+        public static int SqliteClose(IntPtr handle)
+           => _strategy.Close(handle);
+
+        public static string SqliteDbFilename(Sqlite3Handle db, string databaseName)
+            => _strategy.DbFilename(db, databaseName);
+
+        private static readonly StrategyBase _strategy = GetStrategy(new Version(NativeMethods.sqlite3_libversion()));
+
+        private static StrategyBase GetStrategy(Version current)
+        {
+            if (current >= new Version("3.7.15"))
+            {
+                return new Strategy3_7_15();
+            }
+            if (current >= new Version("3.7.14"))
+            {
+                return new Strategy3_7_14();
+            }
+            if (current >= new Version("3.7.10"))
+            {
+                return new Strategy3_7_10();
+            }
+            return new StrategyBase();
+        }
+
+        private class Strategy3_7_15 : Strategy3_7_14
+        {
+            public override string ErrorString(int rc)
+                => NativeMethods.sqlite3_errstr(rc) + " " + base.ErrorString(rc);
+        }
+
+        private class Strategy3_7_14 : Strategy3_7_10
+        {
+            public override int Close(IntPtr handle)
+                => NativeMethods.sqlite3_close_v2(handle);
+        }
+
+        private class Strategy3_7_10 : StrategyBase
+        {
+            public override string DbFilename(Sqlite3Handle db, string databaseName)
+                => NativeMethods.sqlite3_db_filename(db, databaseName);
+        }
+
+        private class StrategyBase
+        {
+            public virtual string ErrorString(int rc)
+                => Strings.DefaultNativeError;
+                
+            public virtual int Close(IntPtr handle)
+                => NativeMethods.sqlite3_close(handle);
+                
+            public virtual string DbFilename(Sqlite3Handle db, string databaseName)
+                => null;
+        }
+    }
+}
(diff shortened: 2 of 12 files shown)


natemcmaster:
cc @bricelam

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
```diff
@@ -60,11 +61,27 @@ public static void ThrowExceptionForRC(int rc, Sqlite3Handle db)
...
-            throw new SqliteException(Strings.FormatSqliteNativeError(rc, message), rc);
+            string message;
+
+            if (db != null && !db.IsInvalid)
+            {
+                message = Strings.FormatSqliteNativeError(
+                            rc,
+                            NativeMethods.sqlite3_errmsg(db));
+            }
+            else if (SqliteVersion.Current >= "3.7.15")
```
bricelam: Careful, the string "3.10" is less than "3.7". May want to use the `Version` class.

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
```diff
@@ -60,11 +61,27 @@ public static void ThrowExceptionForRC(int rc, Sqlite3Handle db)
...
+            {
+                message = Strings.FormatSqliteNativeError(
+                            rc,
+                            NativeMethods.sqlite3_errstr(rc));
```
bricelam: The messages returned by this are so useless. :smile:

src/Microsoft.Data.Sqlite/Utilities/SqliteVersion.cs
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Data.Sqlite.Interop;
+
+namespace Microsoft.Data.Sqlite.Utilities
+{
+    internal class SqliteVersion : IComparable<SqliteVersion>, IComparable
```
bricelam: No need to re-implement `Version`

bricelam:
We should try to put all the switching based on version into its own component.

src/Microsoft.Data.Sqlite/Utilities/SqliteVersion.cs
natemcmaster: Cool. Didn't know this class existed.

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
natemcmaster: SqliteVersion (had) an implicit string conversion operator...too magic?

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
natemcmaster: ¯\_(ツ)_/¯

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
bricelam: Ah yes, too... C++

src/Microsoft.Data.Sqlite/Utilities/SqliteVersion.cs
```diff
@@ -0,0 +1,13 @@
...
+using System;
+using Microsoft.Data.Sqlite.Interop;
+
+namespace Microsoft.Data.Sqlite.Utilities
+{
+    internal class SqliteVersion
+    {
+        public readonly static Version Current = new Version(NativeMethods.sqlite3_libversion());
+    }
+}
```
bricelam: 'Tis a silly class. I like the idea of putting a public static `SQLITE_VERSION` field on `NativeMethods` that gets initialized in the static constructor. `SqliteConnection.ServerVersion` should be updated to use it. Then again, does the value even need to be cached?

src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Data.Sqlite.Interop
+{
+	internal class VersionedMethods
```
natemcmaster: Unified all version-switched calls into one place. @bricelam...name suggestion for this class?

src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
```diff
@@ -0,0 +1,53 @@
...
+            else if (Current >= new Version("3.7.15"))
+            {
+                return Strings.FormatSqliteNativeError(
+                            rc,
+                            NativeMethods.sqlite3_errstr(rc));
+            }
+            return Strings.FormatSqliteNativeErrorStringMissing(rc);
+        }
+
+        internal static bool SqliteClose(IntPtr handle)
```
bricelam: If the class is `internal`, these might as well be `public`.

src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Data.Sqlite.Interop
+{
+    internal class VersionedMethods
```
bricelam: Could take this one step further with the strategy pattern. You'd set the strategy once in the static constructor based on the version then each method wouldn't have to check each time.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -61,9 +61,9 @@ public override string ConnectionString
         public override string Database => MainDatabaseName;
 
         public override string DataSource =>
-            State == ConnectionState.Open
-                ? NativeMethods.sqlite3_db_filename(_db, MainDatabaseName)
-                : ConnectionStringBuilder.DataSource;
+            State == ConnectionState.Open ?
+                VersionedMethods.SqliteDbFilename(_db, MainDatabaseName) ?? ConnectionStringBuilder.DataSource :
+                ConnectionStringBuilder.DataSource;
```
bricelam: My eyeballs dislike these formatting changes :eyes:

bricelam:
:shipit:

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: I'll clean this up. My R# cleanup is broken at the moment.

src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
natemcmaster: Yeah, strategy pattern is a lot cleaner. I updated the implementation.

bricelam: This class is beautiful now. Well done.

PR #172, opened 2015-11-06
Automatically prefix SqliteParameter names during binding

### LINKED ISSUE #80: Don't require parameter name to include prefix
SqlClient allows adding a ParameterName without the prefix, such as @, but Sqlite requires the prefix be included.

Example: this works in SqlClient, but with Sqlite the parameter is not bound to @BronieName

```c#
	var conn = new SqlConnection("Data Source=(localdb)\\MSSQLLocalDB;");
	conn.Open();
	var command = conn.CreateCommand();
	command.CommandText = "SELECT * from Bronies.dbo.Pegasus WHERE Name = @BronieName";
	var parameter = command.CreateParameter();
	parameter.ParameterName="BronieName"; /// <------ this is not required to be @BronieName by SqlClient, but is in Sqlite -->
	parameter.Value="Rainbow Dash 0";
	command.Parameters.Add(parameter);
	command.ExecuteReader()
```

natemcmaster:
Fix #80.

### DIFF
--- a/src/Microsoft.Data.Sqlite/Strings.resx
+++ b/src/Microsoft.Data.Sqlite/Strings.resx
@@ -186,4 +186,7 @@
   <data name="DefaultNativeError" xml:space="preserve">
     <value>For more information on this error code see https://www.sqlite.org/rescode.html</value>
   </data>
+  <data name="AmbiguousParameterName" xml:space="preserve">
+    <value>Cannot bind the value for parameter '{parameterName}' because multiple matching parameters were found in the command text. Specify the parameter name with the symbol prefix, e.g. '@{parameterName}'.</value>
+  </data>
 </root>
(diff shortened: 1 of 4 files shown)


natemcmaster:
cc @sebastienros @bricelam

sebastienros:
Nice, thanks.

src/Microsoft.Data.Sqlite/Strings.resx
```diff
@@ -186,4 +186,7 @@
   <data name="DefaultNativeError" xml:space="preserve">
     <value>For more information on this error code see https://www.sqlite.org/rescode.html</value>
   </data>
+  <data name="AmbiguousParameterName" xml:space="preserve">
+    <value>Cannot bind the value for parameter '{parameterName}' because multiple matching parameters were found in the command text.</value>
```
bricelam: Tell how to fix it. Something like: Specify which one to bind to by including the prefix in the parameter name (e.g. "@{parameterName}").

bricelam:
:shipit:

PR #175, opened 2015-11-11
Add target for easily testing on Linux when working on Windows

natemcmaster:
For local development only. This target does not run on our CI as a part of normal build

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * text=auto
+*.sh text eol=lf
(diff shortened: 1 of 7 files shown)


natemcmaster:
cc @bricelam I'm prototyping this approach here. I want to use a similar layout to run Entity Framework tests.

natemcmaster:
After further consideration and fiddling, this is likely not the best approach. IMO the overhead getting docker working to automate the build is not better than just having a Linux VM and running the project manually inside. Automating linux testing on Windows for this project is overkill.

PR #176, opened 2015-11-13
[small] Better handling for in-memory connection strings

natemcmaster:
Improvements to handle in-memory connection strings. 

`SqlConnection.DataSource` returned differencly based on whether the connection is open or not. Instead of relying on sqlite3_db_filename, this API now includes the logic used to adjust for relative paths.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,4 @@ project.lock.json
 .settings
 *.sublime-*
 *.pfx
-
-# Ignore vagrant - used for cross-platform testing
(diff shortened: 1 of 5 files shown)


bricelam:
I don't like this; it feels like we're guessing. `sqlite3_db_filename` defers to the [VFS](https://www.sqlite.org/vfs.html) which is a black box that we shouldn't make any assumptions about...

natemcmaster:
Take 2. #181

PR #181, opened 2015-12-02
Improved support for in-memory databases

natemcmaster:
Support connection strings like `file:memdb1?mode=cached`

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteCommand.cs
@@ -122,6 +122,10 @@ namespace Microsoft.Data.Sqlite
 
             do
             {
+                var currentStatement = tail.Split(';')[0];
+                var isReadOnlyPragma = currentStatement.StartsWith("PRAGMA ", StringComparison.OrdinalIgnoreCase)
+                                       && !currentStatement.Contains("=");
+
                 Sqlite3StmtHandle stmt;
                 var rc = NativeMethods.sqlite3_prepare_v2(
                         Connection.DbHandle,
@@ -187,9 +191,9 @@ namespace Microsoft.Data.Sqlite
                     throw;
                 }
 
-                // NB: This is only a heuristic to separate SELECT statements from INSERT/UPDATE/DELETE statements. It will result
+                // NB: This is only a heuristic to separate PRAGMA and SELECT statements from INSERT/UPDATE/DELETE statements. It will result
                 //     in unexpected corner cases, but it's the best we can do without re-parsing SQL
-                if (NativeMethods.sqlite3_stmt_readonly(stmt) != 0)
+                if (NativeMethods.sqlite3_stmt_readonly(stmt) != 0 || isReadOnlyPragma)
                 {
                     stmts.Enqueue(Tuple.Create(stmt, rc != SQLITE_DONE));
                 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -122,6 +122,7 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
 
             do
             {
+                var isPragma = tail.StartsWith("PRAGMA ", StringComparison.OrdinalIgnoreCase);
```
bricelam: Too brittle. Is this needed?

natemcmaster: `"PRAGMA journal_mode;"` throws with out this.

```
System.ArgumentNullException: Parameter name: SafeHandle cannot be null.
      Stack Trace:
           at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
           at Microsoft.Data.Sqlite.Interop.NativeMethods.sqlite3_column_type(Sqlite3StmtHandle stmt, Int32 iCol)
           at Microsoft.Data.Sqlite.SqliteDataReader.GetSqliteType(Int32 ordinal)
```

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -94,12 +111,12 @@ public override void Open()
...
             }
 
-            var flags = Constants.SQLITE_OPEN_READWRITE | Constants.SQLITE_OPEN_CREATE;
+            var flags = Constants.SQLITE_OPEN_READWRITE | Constants.SQLITE_OPEN_CREATE | Constants.SQLITE_OPEN_URI;
             flags |= (ConnectionStringBuilder.Cache == SqliteConnectionCacheMode.Shared) ? Constants.SQLITE_OPEN_SHAREDCACHE : Constants.SQLITE_OPEN_PRIVATECACHE;
 
             var path = ConnectionStringBuilder.DataSource;
 
-            if (!path.Equals(":memory:", StringComparison.OrdinalIgnoreCase))
+            if (!IsInMemory(path))
```
bricelam: I want to provide some better abstractions in this area. Connection strings that look like `Data Source=file:thedb?mode=memory; Cache=Shared;` are terrible.

src/Microsoft.Data.Sqlite/SqliteCommand.cs
bricelam: Ah, looks like we have another heuristic we can use. :smile:

src/Microsoft.Data.Sqlite/SqliteConnection.cs
natemcmaster: So, instead receive ConnectionStringBuilder params and translate to their equivalent `SQLITE_OPEN_*` [flags](https://www.sqlite.org/c3ref/c_open_autoproxy.html)?

src/Microsoft.Data.Sqlite/SqliteCommand.cs
natemcmaster: Updated to only match true "readonly" pragmas. Better?

natemcmaster:
Ping. Even if we do add a first-class connection string parameter for "inmemory", we still want to support URI syntax, right?

PR #186, opened 2015-12-03
Enable using winsqlite3.dll

### LINKED ISSUE #179: Enable using Windows in-the-box version of SQLite
Windows is shipping a version of SQLite in the box. We should look into enabling our library to use it.

bricelam:
To use, call `SqliteEngine.UseWinSqlite3()` before using SQLite.

Fixes #179

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
@@ -7,67 +7,65 @@ namespace Microsoft.Data.Sqlite.Interop
 {
     internal class VersionedMethods
     {
-        public static string SqliteErrorMessage(int rc, Sqlite3Handle db)
-        {
-            var message = db == null || db.IsInvalid
-                ? _strategy.ErrorString(rc)
-                : NativeMethods.sqlite3_errmsg(db);
+        private static readonly BaseStrategy _strategy;
 
-            return Strings.FormatSqliteNativeError(rc, message);
+        static VersionedMethods()
+        {
+            var version = new Version(NativeMethods.sqlite3_libversion());
+            if (version >= new Version(3, 7, 15))
+            {
+                _strategy = new Strategy3_7_15();
+            }
+            else if (version >= new Version(3, 7, 14))
+            {
+                _strategy = new Strategy3_7_14();
+            }
+            else if (version >= new Version(3, 7, 10))
+            {
+                _strategy = new Strategy3_7_10();
+            }
+            else
+            {
+                _strategy = new BaseStrategy();
+            }
         }
 
-        public static int SqliteClose(IntPtr handle)
-           => _strategy.Close(handle);
+        public static string GetErrorString(int rc)
+            => _strategy.GetErrorString(rc);
 
-        public static string SqliteDbFilename(Sqlite3Handle db, string databaseName)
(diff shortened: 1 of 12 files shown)


src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
```diff
@@ -7,67 +7,56 @@ namespace Microsoft.Data.Sqlite.Interop
...
-            var message = db == null || db.IsInvalid
-                ? _strategy.ErrorString(rc)
-                : NativeMethods.sqlite3_errmsg(db);
+        private static readonly BaseStrategy _strategy;
 
-            return Strings.FormatSqliteNativeError(rc, message);
+        static VersionedMethods()
+        {
+            var version = new Version(NativeMethods.sqlite3_libversion());
+            _strategy = version >= new Version(3, 7, 15)
```
natemcmaster: IMO the `if` statements are more readable than multi-nested ternary operators...but this is not a blocking objection.

src/Microsoft.Data.Sqlite/SqliteEngine.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Data.Sqlite.Interop;
+
+namespace Microsoft.Data.Sqlite
+{
+    public static class SqliteEngine
```
natemcmaster: Could we also add "UseDefault" to revert the engine back to `"sqlite3"`?

src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs
bricelam: Will revert. I got a little carried away with renames and simplifications. The if statements did look better.

natemcmaster:
:shipit:

src/Microsoft.Data.Sqlite/SqliteEngine.cs
bricelam: I'll bring it up in a design meeting. It's an awkward API all around.

natemcmaster: Could this be a connection string parameter instead?

bricelam: Also note, you can't change it after it's loaded. If you don't want to use it, don't call it. Having another API to un-call it is silly.

bricelam: > Could this be a connection string parameter instead?

No, it's not per-connection. It's "global" (AppDomain) for the entire library.

natemcmaster: > No, it's not per-connection

...currently. But can't we dynamically switch to a different `ISqlite3` implementation?

Connection string seems like natural place to put this kind of configuration. Ideally the setting (something like`UseWinsqlite3=True;`) could handle the "if windows 10 do this, else do that"  for users so we wouldn't need different connection strings for Linux, Windows 7 and Windows 10.

bricelam: Could you file a bug? I agree it could be useful.

PR #192, opened 2015-12-08
Add Mode connection string option

### LINKED ISSUE #180: Remove hard dependency on Microsoft.Extensions.PlatformAbstractions for net45
Resolving relative paths can be done using `AppContext`.
DNX runtime detection, can try and dynamically load the assembly at runtime.

### LINKED ISSUE #182: Support named, in-memory databases
The native API supports this using connection strings like `file:cow-db?mode=memory&cache=shared`

### LINKED ISSUE #188: Relative path in connection string pointing to different locations
I'm using a two project setup as described here https://github.com/aspnet/Microsoft.Data.Sqlite/issues/55#issuecomment-132480553

config.json
````
    "Data": {
        "DefaultConnection": {
            "ConnectionString": "Filename=database.db"
        }
    },
````

Startup.cs
````
[...]
            var builder = new ConfigurationBuilder()
                 .SetBasePath(appEnv.ApplicationBasePath)
                 .AddJsonFile("config.json");
[...]

            services.AddEntityFramework()
                .AddSqlite()
                .AddDbContext<MyAppDbContext>(
                    options => options.UseSqlite(_configuration["Data:DefaultConnection:ConnectionString"]));

[...]
````
I have a custom command defined in the Program.cs allowing me to reset and reseed the database.

````
    private static void ResetDatabase()
    {
        var dbContext = _services.GetService<MyAppDbContext>();

        dbContext.Database.EnsureDeleted();

        dbContext.Database.EnsureCreated();
        
        Console.WriteLine("Database reset successful.");
    }
````

When I run my application from IIS Express the database gets created in :
`D:\MyApp\src\MyApp\wwwroot`

When I run the custom command it appears in:
`D:\MyApp\src\MyApp\`

### PR DESCRIPTION
**Pull request by bricelam**
This also...
* Removes the PlatformAbstractions dependency (resolves #180)
* Fixes #188

Resolves #182

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnection.cs
@@ -4,14 +4,16 @@
 using System;
 using System.Data;
 using System.Data.Common;
+using System.Diagnostics;
 using System.IO;
+using Microsoft.Data.Sqlite.Interop;
+
 #if NETCORE50
 using System.Reflection;
 using Microsoft.Data.Sqlite.Utilities;
-#else
-using Microsoft.Extensions.PlatformAbstractions;
 #endif
-using Microsoft.Data.Sqlite.Interop;
+
+using static Microsoft.Data.Sqlite.Interop.Constants;
 
 namespace Microsoft.Data.Sqlite
 {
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: src/Microsoft.Data.Sqlite/SqliteConnection.cs [hunk 1]
```diff
@@ -103,89 +104,103 @@ public override void Open()
...
+                    break;
+
+                case SqliteCacheMode.Private:
+                    flags |= SQLITE_OPEN_PRIVATECACHE;
+                    break;
+
+                default:
+                    Debug.Assert(
+                        ConnectionStringBuilder.Cache == SqliteCacheMode.Default,
+                        "ConnectionStringBuilder.Cache is not Unspecified.");
```
natemcmaster: `not Unspecified`? Not sure what this means.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -103,89 +104,103 @@ public override void Open()
...
+
+            OnOpened();
+        }
+
+        partial void OnOpened();
+
+#if NETCORE50
+        partial void OnOpened()
+        {
+            var temporaryFolder = CurrentApplicationData?.TemporaryFolder.Path;
```
natemcmaster: I believe we replace this with `Path.GetTempPath()`.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
```diff
@@ -81,7 +81,17 @@ public static bool IsLoaded(string dllName)
         }
 
         private static bool IsDNX()
-            => PlatformServices.Default?.Application.RuntimeFramework.Identifier == "DNX";
+        {
+            dynamic defaultPlatformServices =
```
natemcmaster: Clever hack. Wish there were a better way cuz this feels gross.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
```diff
@@ -175,6 +155,29 @@ public void Open_works()
         }
 
         [Fact]
+        public void Can_use_named_shareable_inMemory_databases()
```
natemcmaster: Would `Data Source=file:people?mode=memory&cache=shared` also work?

src/Microsoft.Data.Sqlite/SqliteConnection.cs
bricelam: Oops, I renamed it to Default, but missed this assert.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
bricelam: Yep, you can always "drop down" to using URIs. This just shows the "supported" ADO.NETty pattern.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
bricelam: We may be able to just use `AppDomain.CurrentDomain.SetupInformation.TargetFrameworkName`. I'll investigate.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
bricelam: Will try. Much simpler.

bricelam: Hmm... They're slightly different directories. The SQLite documentation uses `ApplicationData.Current.TemporaryFolder.Path`. We'll stick with that for now.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
bricelam: No good. Too many nuances. We'll stick with this hack for now. This code path may go away with CLI.

natemcmaster:
The only thing I see missing is tests for actually opening URI connection strings. Are they already in the tests somewhere? Otherwise, :shipit:

bricelam:
lol, ok, ok, I'll add more tests... I just got lazy. :sheep:

bricelam:
Added missing tests for ReadOnly, ReadWrite, and URIs.

natemcmaster:
:ship: :it:

PR #198, opened 2015-12-11
Workaround TD.net bug

natemcmaster:
see https://github.com/aspnet/PlatformAbstractions/issues/2

cc @bricelam


bricelam:
I might just wrap the whole method in a `try..catch`. If anything goes wrong, assume it's not DNX.

PR #199, opened 2015-12-12
Replace call to Windows.Storage via DLR with reflection calls.

### LINKED ISSUE #190: InvalidOperationException  from AdjustForRelativeDirectory
**Copied from https://github.com/aspnet/EntityFramework/issues/3935. Per the comments, this is an exception from our reflection code in `Microsoft.Data.Sqlite.dll!Microsoft.Data.Sqlite.SqliteConnection.AdjustForRelativeDirectory(string path)`.**

<hr />

I have an issue similar to #2184. Context with simple model fails on any usage. On Database.Migrate(), trying to use DbSet, etc.

# Model

    public class Book
    {
        public int Id { get; set; }
        public string Title { get; set; }
        public bool Done { get; set; }
    }

    public class Movie
    {
        public int Id { get; set; }
        public string Title { get; set; }
        public bool Done { get; set; }
    }

    public class Context : DbContext
    {
        public DbSet<Book> Books { get; set; }
        public DbSet<Movie> Movies { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlite("Filename=db.db");
        }
    }

# App.xaml.cs

It's almost empty!

    public sealed partial class App
    {
        public App()
        {
            InitializeComponent();

            using (var db = new Context())
            {
                db.Database.Migrate(); // exception occures here
            }
        }
    }

# Error

> System.InvalidOperationException occurred
  HResult=-2146233079
  Message=The API 'System.Reflection.MemberInfo.get_MetadataToken()' cannot be used on the current platform. See http://go.microsoft.com/fwlink/?LinkId=248273 for more information.
  Source=mscorlib
  StackTrace:
       at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)

natemcmaster:
Fix #190

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnection.cs
@@ -179,20 +179,29 @@ namespace Microsoft.Data.Sqlite
 #if NETCORE50
         partial void OnOpened()
         {
-            var temporaryFolder = CurrentApplicationData?.TemporaryFolder.Path;
-            if (temporaryFolder != null)
+            var appDataType = CurrentApplicationData?.GetType();
+            var temporaryFolder = appDataType?.GetRuntimeProperty("TemporaryFolder").GetValue(CurrentApplicationData);
+            var temporaryFolderPath = temporaryFolder?.GetType().GetRuntimeProperty("Path").GetValue(temporaryFolder) as string;
+            if (temporaryFolderPath != null)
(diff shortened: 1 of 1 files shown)


bricelam:
:shipit:

PR #201, opened 2015-12-17
Fix build warning on .NET Native

natemcmaster:
Using "ExactSpelling" suppresses dozens of build warnings on UWP because winsqlite3.dll does not exist in .NET Native.


bricelam:
Oh weird; I thought that only made a difference on methods with strings in their definition.

Can you change the `*.tt` file used to generate this `*.cs` file too?

natemcmaster:
[Here is the *.tt change](https://github.com/aspnet/Microsoft.Data.Sqlite/pull/201/files#diff-bb5a632a9ceb315ec42e52d7bfa0e77d)

>  I thought that only made a difference on methods with strings in their definition.

On .NET Native, ILC compiles p/invokes into native calls AOT. Setting `ExactSpelling = true` tells ILC that *"I know winsqlite3.dll does not exist, but I know what I'm doing so don't bother me."*

bricelam:
:shipit:

Issue #205, opened 2015-12-17
Support SQLite's transaction types


natemcmaster:
Sqlite supports three different transaction types:

 - `DEFERRED` (default)
 - `IMMEDIATE`
 - `EXCLUSIVE`

See https://www.sqlite.org/lang_transaction.html.

Currently `SqliteTransaction` only exposes the default (deferred)

natemcmaster:
Using immediate mode can help prevent SQLITE_BUSY errors in the middle of a transaction.

PR #207, opened 2015-12-22
Support SQLite transaction types

natemcmaster:
Fix https://github.com/aspnet/Microsoft.Data.Sqlite/issues/205

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnection.cs
@@ -238,10 +238,17 @@ namespace Microsoft.Data.Sqlite
         // NB: Other providers don't set Transaction
         public new virtual SqliteCommand CreateCommand() => new SqliteCommand { Connection = this, Transaction = Transaction };
         protected override DbCommand CreateDbCommand() => CreateCommand();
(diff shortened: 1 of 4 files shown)


bricelam:
@divega Are there similar patterns on other providers that we should be following?

bricelam:
I dug into this a bit more, and I think we have a bug around `IsolationLevel.Serialzable`. Here is how I think the isolation levels should be translated to SQLite.

IsolationLevel | SQL
-------------- | ---
ReadCommitted | PRAGMA read_uncommitted=0; BEGIN;
ReadUnommitted | PRAGMA read_uncommitted=1; BEGIN;
Serializable | PRAGMA read_uncommitted=0; BEGIN IMMEDIATE;

I'm not sure how `BEGIN EXCLUSIVE` maps (possibly `RepeatableRead`), but we can punt on this.

natemcmaster:
Would be nice to have a test for the bug. Using internal email powers to see if we can reproduce the potential bug (and the fix) in IsolationLevel.Serializable

bricelam:
I also started writing some functional tests in [bricelam:txtst](https://github.com/aspnet/Microsoft.Data.Sqlite/compare/dev...bricelam:txtst) to make sure we're not braking the `IsolationLevel` semantics.

natemcmaster:
Close in favor of #213

PR #208, opened 2015-12-22
Return empty byte array

### LINKED ISSUE #203: Empty blob should return byte[0]
A column with type BLOB and length = 0 should return `byte[0]` and `IsDBNull == false`.

Instead, SqliteDataReader returns `null` and `IsDBNull == false`.

Repro: https://gist.github.com/natemcmaster/6f2ef57338c597ccfa65

natemcmaster:
fix #203

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
@@ -86,12 +86,15 @@ namespace Microsoft.Data.Sqlite.Interop
         public static int sqlite3_close_v2(IntPtr db)
             => Sqlite3.close_v2(db);
 
+        // TODO can be Array.Empty<T>() when upgrading to net46
+        private static byte[] EmptyByteArray = new byte[0];
+
         public static byte[] sqlite3_column_blob(Sqlite3StmtHandle pStmt, int iCol)
         {
             var ptr = Sqlite3.column_blob(pStmt, iCol);
             if (ptr == IntPtr.Zero)
             {
-                return null;
+                return EmptyByteArray;
             }
 
             var bytes = Sqlite3.column_bytes(pStmt, iCol);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -91,7 +91,7 @@ public static byte[] sqlite3_column_blob(Sqlite3StmtHandle pStmt, int iCol)
             var ptr = Sqlite3.column_blob(pStmt, iCol);
             if (ptr == IntPtr.Zero)
             {
-                return null;
+                return new byte[0];
```
davidfowl: `Array.Empty<byte>()`

natemcmaster: `.Empty<T>()` doesn't exist .net < 4.6 (Mono)

davidfowl: ifdef?

bricelam:
Does similar behavior occur for empty strings? Are NULL and an empty array the same in C? Why is the native API returning NULL?

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
bricelam: Boo. I hate #if-in just to use Array.Empty. If caching is really an issue, cache it statically on this type.

natemcmaster:
> Strings returned by sqlite3_column_text() and sqlite3_column_text16(), even empty strings, are always zero-terminated. The return value from sqlite3_column_blob() for a zero-length BLOB is a NULL pointer.

https://www.sqlite.org/c3ref/column_blob.html.

[Zero-length arrays](https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html) in C depend on the version of C.

bricelam:
:shipit: Thanks for following up.

bricelam:
Does `SqliteDataReader.IsDBNull` return `true` For these values?

natemcmaster:
This is the original issue. This fixes this misalignment of SqliteDataReader returning `null` even though `IsDBNull == false`.

PR #209, opened 2015-12-22
Allow parameter binding to null

### LINKED ISSUE #200: Allow parameter binding to clr null
`SqliteParameter` throws if _value is `null`.

https://github.com/aspnet/Microsoft.Data.Sqlite/blob/85403c053c2c1d5bce4ea03d310ee506d4a2653d/src/Microsoft.Data.Sqlite/SqliteParameter.cs#L143

Consider instead calling `NativeMethods.sqlite3_bind_null` if the clr value is null. Investigate the potential implications on null semantics.

natemcmaster:
Fix #200

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteParameter.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteParameter.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Data.Sqlite
         private string _parameterName = string.Empty;
         private object _value;
         private Action<Sqlite3StmtHandle, int> _bindAction;
+        private bool _unset = true;
         private bool _bindActionValid;
 
         public SqliteParameter()
@@ -111,6 +112,7 @@ namespace Microsoft.Data.Sqlite
             set
             {
                 _value = value;
+                _unset = false;
                 _bindActionValid = false;
             }
         }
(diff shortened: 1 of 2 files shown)


bricelam:
Hmm, I thought I mimicked the behavior of other providers. Was this just a quirk of SQLite.NET? If so, I don't think we should do it. DBNull is weird, but this is how I understood ADO.NET providers should be implemented.

natemcmaster:
SqlClient doesn't allow `null`. I haven't tested others. Likely a quirk of SQLite.NET (there are many such quirks.)

ErikEJ:
@natemcmaster sqlite.net is not an ado.net provider. Do you mean System.Data.Sqlite ?

natemcmaster:
Confirmed that this change matches the behavior SQLite.NET *and* System.Data.Sqlite. Parameter value of `null` AND `DBNull.Value` is accepted in both.

bricelam:
I vote for not merging since, IMHO, it breaks the ADO.NET contract where `null` means unset, not `DbNull`.

natemcmaster:
I'm reverting my thinking on this now that I realize this is the ADO contract, even though I suspect users will think this is a bug if they switch from existing SQLite ADO implementations. I'm going to improve the error message so users understand they need to use DBNull.

Issue #204, opened 2015-12-17
SqliteDataReader.GetFieldType returns Int32 if there are no rows.


simonkang:
This is related to Issue #202 

I am using Dapper and it caches the field types by sql. If the first time the sql is executed all of the columns' types in the result set are cached as Int32. Then when there is data, cast exceptions can occur when the actual data is Int64.

The issue is related to using sqlite3_column_type. If sqlite3_column_decltype can be used, then inferred ClrType is returned.

PR #210, opened 2015-12-22
Use Declared Table Types in Data Reader with sqlite3_column_decltype

simonkang:
For GetFieldType and GetFieldValue, use the column type from
sqlite3_column_decltype to determine the .NET type to use / convert.

This is to resolve issue #204. I want to point out the last line in GetFieldValue. I have changed it to throw an Exception if the type is not handled, which is a change from "base.GetFieldValue<T>(ordinal)".

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConvert.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConvert.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.Sqlite
+{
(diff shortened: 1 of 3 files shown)


bricelam:
We use to have code like this, but we ripped it all out in favor of not interpreting the column type name since they have virtually no semantics in SQLite. I believe all of this logic could be layered over the existing implementation.

simonkang:
That does make sense. I looked through the history and it looks like it got removed with issue #68. I will try your suggestion and use GetDataTypeName to override the GetFieldType implementation.

simonkang:
Thank you for the suggestion. It works well. I basically wrapped the Sqlite objects with a custom DbConnection, DbCommand, and DbDataReader objects and changed the implementation of the GetFieldType.

PR #211, opened 2015-12-23
Improve error message for common misuse of SqliteParameter

natemcmaster:
See #209


natemcmaster:
Ping on this old PR @bricelam

bricelam:
This feels unnecessary to me. We're honoring the ADO.NET contract, I don't think we need to defend that because SQLite.NET didn't.

divega:
I can go either way re having a custom exception for this, but I like the `nameof()` change :smile:

natemcmaster:
Going over old PRs...can't remember the original motivation for this anymore.

PR #216, opened 2016-01-23
Swapped IsDNX and IsMono to fix running Mono Mobile(Xamarin.Android/Xamarin.iOS)

DavidKarlas:
Crash happens when IsDNX method is loaded because it tries to load System.Runtime.Versioning.FrameworkName type which is not present on Mono Mobile. If IsMono is called first IsDNX is never called, hence no crash.


davidfowl:
This is too subtle to not write a giant comment saying why the ordering of conditions within the if statement matter. Why doesn't the try catch work on iOS and Android?

DavidKarlas:
Added comment.

Reason why try catch is not working is because exception happens while constructing method and not finding FrameworkName type, so before entering try catch statement.

Added new PR that you might like more:
https://github.com/aspnet/Microsoft.Data.Sqlite/pull/217

bricelam:
This PR has been superseded by #225

PR #217, opened 2016-01-24
Removed statically referencing FrameworkName which is missing in Mono Mobile.

DavidKarlas:
With this change if runtime is missing System.Runtime.Versioning.FrameworkName it doesn't crash during method loading.


bricelam:
This PR has been superseded by #225

PR #223, opened 2016-02-02
Fix non-Windows dotnet build

natemcmaster:
cc @victorhurdugaci @bricelam 

KoreBuild tries to restore/build our test UWP app...does not work on xplat for obivous reasons.


bricelam:
Would renaming the `project.json` to `Microsoft.Data.Sqlite.FunctionalTests.UWP.project.json` work?

natemcmaster:
Yup, let's do that. I didn't know VS tooling would still recognize `<project name>.project.json`.

bricelam:
Oh what knowledge you can gain from spelunking MSBuild scripts! :wink:

bricelam:
:shipit:

natemcmaster:
Depends on https://github.com/aspnet/Universe/pull/365

victorhurdugaci:
This will also be fixable when this goes in https://github.com/aspnet/Universe/issues/360

victorhurdugaci:
But that should be okay as long as your build system understands different names other than `project.json`

natemcmaster:
In this case, we want CI build system to skip the UWP test project but VS to still work.

PR #225, opened 2016-02-09
Test using .NET Core CLI

### LINKED ISSUE #218: Review NativeLibraryLoader from a .NET Core CLI perspective
Now that DNX is gone, we should review the implementation of NativeLibraryLoader to ensure it still makes sense.

bricelam:
Also resolves #218

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteDataReader.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteDataReader.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Data.Sqlite
 #if NET451
             return new DbEnumerator(this);
 #else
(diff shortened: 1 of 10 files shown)


src/Microsoft.Data.Sqlite/SqliteDataReader.cs
```diff
@@ -83,7 +83,7 @@ public override IEnumerator GetEnumerator()
 #if NET451
             return new DbEnumerator(this);
 #else
-    // TODO: Remove when the System.Data.Common includes DbEnumerator
+            // TODO: Remove when the System.Data.Common includes DbEnumerator
```
divega: I think DbEnumerator was already there when I checked las week.

divega:
:shipit:

PR #231, opened 2016-02-20
Make the first and only call to set temp_store_directory thread safe

### LINKED ISSUE #229: Do not use deprecated PRAGMA temp_store_directory
[SqliteConnection.cs:187](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/SqliteConnection.cs#L187) uses `PRAGMA temp_store_directory` which is [deprecated from SQLite](http://sqlite.org/pragma.html#pragma_temp_store_directory)

Another potential issue with using that pragma is that it must be used only once and its not thread safe, quote from SQLite docs:

>  If the temp_store_directory is changed for one database connection while other database connections are open in the same process, then the behavior is undefined and probably undesirable...

> ...Changing the temp_store_directory setting is not threadsafe

It's causing crashes in our UWP project when SQLite is used from multiple threads.

natemcmaster:
Fix #229 

cc @bricelam

### DIFF
--- a/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
+++ b/src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
@@ -3,6 +3,9 @@
 
 using System;
 using System.Runtime.InteropServices;
+#if !NET451
+using Microsoft.Data.Sqlite.Utilities;
+#endif
 
 namespace Microsoft.Data.Sqlite.Interop
 {
@@ -21,7 +24,41 @@ namespace Microsoft.Data.Sqlite.Interop
             _dllName = dllName;
         }
 
-        private static ISqlite3 Sqlite3 => _sqlite3.Value;
+        private static ISqlite3 Sqlite3
+        {
+            get
+            {
+                if (!_sqlite3.IsValueCreated)
+                {
+                    OnLoad();
+                }
+                return _sqlite3.Value;
+            }
+        }
+
+        static partial void OnLoad();
+
+#if !NET451
+        private static volatile bool _isLoaded;
+
+        static partial void OnLoad()
+        {
+            if (_isLoaded)
+            {
+                // prevent re-entrant calls
+                return;
+            }
+            _isLoaded = true;
+            if (ApplicationDataHelper.TemporaryFolderPath != null)
+            {
+                using (var connection = new SqliteConnection("Data Source=:memory:"))
+                {
+                    connection.Open();
+                    connection.ExecuteNonQuery("PRAGMA temp_store_directory = '" + ApplicationDataHelper.TemporaryFolderPath + "';");
+                }
+            }
+        }
+#endif
 
         public static int sqlite3_bind_blob(Sqlite3StmtHandle pStmt, int i, byte[] zData, int nData, IntPtr xDel)
             => Sqlite3.bind_blob(pStmt, i, zData, nData, xDel);
--- a/src/Microsoft.Data.Sqlite/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnection.cs
@@ -9,7 +9,6 @@ using System.IO;
 using Microsoft.Data.Sqlite.Interop;
 
 #if !NET451
-using System.Reflection;
 using Microsoft.Data.Sqlite.Utilities;
 #endif
 
@@ -105,8 +104,6 @@ namespace Microsoft.Data.Sqlite
                 throw new InvalidOperationException(Strings.OpenRequiresSetConnectionString);
             }
 
-
-
             var filename = ConnectionStringBuilder.DataSource;
             var flags = 0;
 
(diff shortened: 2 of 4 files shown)


bricelam:
Hmm, I don't like that we're doing this when a connection is first to open, and I don't like all the orchestration. Can we connect to `:memory:` and do it in `NativeMethods.Load()`? That method is guaranteed to only run once per `AppDomain`. It would also guarantee that the the temp directory gets set for paths that might need it before `sqlite3_open_v2` is called.

natemcmaster:
Clever idea. Updated to use this simpler approach.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -2,12 +2,34 @@
...
 using System;
+using System.Reflection;
 using System.Runtime.InteropServices;
+using Microsoft.Data.Sqlite.Utilities;
 
 namespace Microsoft.Data.Sqlite.Interop
 {
     internal static partial class NativeMethods
     {
+        static NativeMethods()
```
bricelam: Can't do it here. This will cause the native assembly to be loaded and `SqliteEngine.UseWinSqlite3` will always throw.

src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -170,25 +168,10 @@ public override void Open()
...
-            var temporaryFolder = appDataType?.GetRuntimeProperty("TemporaryFolder").GetValue(CurrentApplicationData);
-            var temporaryFolderPath = temporaryFolder?.GetType().GetRuntimeProperty("Path").GetValue(temporaryFolder) as string;
-            if (temporaryFolderPath != null)
-            {
-                DbConnectionExtensions.ExecuteNonQuery(this, "PRAGMA temp_store_directory = '" + temporaryFolderPath + "';");
-            }
-        }
-
-        private static object CurrentApplicationData
+        internal static object CurrentApplicationData
```
bricelam: Instead of making this property internal, we should probably move it into its own helper/utility class.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
bricelam: Do it inside `Load`--immediately after the native library is loaded.

natemcmaster: I'm not sure where we can put it then w/o causing a recursive call.

bricelam: :smirk: We could call the native APIs directly...
Or just use a flag to detect re-entrance and no-op.

natemcmaster:
So FWIW I've tried the static constructor idea in a few places and it doesn't work for a variety of complications. Inside Load, just after Load, etc. They all have issues. I'm reverting this PR back to the original approach until I find something simpler that actually works.

natemcmaster:
Updated. Unfortunately I can't actually test this on UWP since .NET Core for UWP hasn't updated lately.

natemcmaster:
Ping for review

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -21,7 +24,41 @@ public static void SetDllName(string dllName)
...
+                    OnLoad();
+                }
+                return _sqlite3.Value;
+            }
+        }
+
+        static partial void OnLoad();
+
+#if !NET451
+        private static bool _isLoaded;
```
bricelam: Should probably be `volatile`

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -21,7 +24,41 @@ public static void SetDllName(string dllName)
...
 
-        private static ISqlite3 Sqlite3 => _sqlite3.Value;
+        private static ISqlite3 Sqlite3
+        {
+            get
+            {
+                if (!_sqlite3.IsValueCreated)
+                {
+                    OnLoad();
+                }
```
bricelam: I would *really* like to try and get this logic inside the `Load` method.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -21,7 +24,41 @@ public static void SetDllName(string dllName)
...
+                return;
+            }
+            _isLoaded = true;
+            if (ApplicationDataHelper.TemporaryFolderPath != null)
+            {
+                using (var connection = new SqliteConnection("Data Source=:memory:"))
+                {
+                    connection.Open();
+                    connection.ExecuteNonQuery("PRAGMA temp_store_directory = '" + ApplicationDataHelper.TemporaryFolderPath + "';");
+                }
```
bricelam: The more I think about this, the more I think we should make native calls directly to do all of this.

bricelam:
:shipit: But file an issue for someone to address my comments if you want to punt on them.

natemcmaster:
Filed https://github.com/aspnet/Microsoft.Data.Sqlite/issues/240

Issue #222, opened 2016-01-28
Concurrency issues; possibly support connection pooling or setting default timeout


gingters:
When you have two threads that SELECT some data and then INSERT or UPDATE a row on the same table on the same time, a "Database is locked" error is thrown.

This happens on Windows as well as on Mac OS X (was not yet tested on linux).

My approach would be to solve this by enabling connection pooling (which Microsoft.Data.Sqlite does not support) or possibly set a larger command timeout (which Microsoft.Data.Sqlite does not support on the connection string or on the connection itself). Since EF owns the actual command its not possible to set another timeout value there, too.

Please find attached a project that reproduced this issue :
[Sqlite.Repro.zip](https://github.com/aspnet/Microsoft.Data.Sqlite/files/107937/Sqlite.Repro.zip)

Simply execute "dnu restore && dnx run" to experience the issue.

natemcmaster:
This looks like it could be the same serializable bug we fixed in https://github.com/aspnet/Microsoft.Data.Sqlite/pull/213 (which is in build >= 15974). @gingters does your issue still occur with this fix?

rowanmiller:
Closing as we have not heard back

natemcmaster:
I found a similar, related issue for which I filed a fix in #232. This and #213 could have been the cause of the "database is locked" exceptions.

PR #232, opened 2016-02-22
Fix for SQLITE_BUSY errors throw before command timeout expired

natemcmaster:
Can occur in multi-threaded context. This fixes failing concurrency tests that started failing on OSX and Linux.

Possibly the cause of #222

cc @bricelam


bricelam:
:shipit:

PR #238, opened 2016-03-11
Add overrides for default base directory

natemcmaster:
cc @bricelam

### DIFF
--- a/src/Microsoft.Data.Sqlite/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite/SqliteConnection.cs
@@ -211,13 +211,15 @@ namespace Microsoft.Data.Sqlite
             {
                 var appDataType = CurrentApplicationData?.GetType();
                 var localFolder = appDataType?.GetRuntimeProperty("LocalFolder").GetValue(CurrentApplicationData);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Data.Sqlite/SqliteConnection.cs
```diff
@@ -211,7 +211,8 @@ private static string BaseDirectory
             {
                 var appDataType = CurrentApplicationData?.GetType();
                 var localFolder = appDataType?.GetRuntimeProperty("LocalFolder").GetValue(CurrentApplicationData);
-                return (localFolder?.GetType().GetRuntimeProperty("Path").GetValue(localFolder) as string)
+                return Environment.GetEnvironmentVariable("ADONET_DATA_DIR")
```
bricelam: We need this on `net451` too (a few lines down)

natemcmaster: Updated.

bricelam:
:shipit:

davidfowl:
Can this be settable via a first class API as well?

bricelam:
@davidfowl Wouldn't you just use a full path if you wanted that?

natemcmaster:
This is really just a workaround for https://github.com/aspnet/EntityFramework/issues/4765, not really for general use.

PR #242, opened 2016-04-09
Cross-compile to netcore50 to support EF on UWP

natemcmaster:
Part of https://github.com/aspnet/EntityFramework/pull/5021

### DIFF
--- a/src/Microsoft.Data.Sqlite/project.json
+++ b/src/Microsoft.Data.Sqlite/project.json
@@ -13,6 +13,9 @@
     "net451": {
       "frameworkAssemblies": {
         "System.Data": ""
+      },
+      "runtimes": {
+        "any": {}
       }
     },
     "netstandard1.3": {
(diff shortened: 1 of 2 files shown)


davidfowl:
Why is there a runtimes section?

natemcmaster:
@davidfowl it was a workaround for a bug in dotnet-pack for a previous draft of this PR and I forgot to check if I could remove it. It appears this works without so i'll take it out.

src/Microsoft.Data.Sqlite/project.json
```diff
@@ -27,6 +27,31 @@
...
+        "System.Reflection.Extensions": "4.0.0",
+        "System.Resources.ResourceManager": "4.0.0",
+        "System.Runtime.Extensions": "4.0.10",
+        "System.Runtime.InteropServices": "4.0.20",
+        "System.Text.Encoding": "4.0.10",
+        "System.Threading": {
+          "type": "build",
+          "version": "4.0.10"
+        },
+        "System.Text.RegularExpressions": {
```
davidfowl: Why would you need build time dependencies?

natemcmaster: This, and all other build time dependencies I added, are only there so restore will pass.

natemcmaster:
@davidfowl we have ask-mode permission to merge. Even though we all agree -- cross-compiling is an unfortunate workaround -- is this change okay to merge?

PR #250, opened 2016-05-16
Use external sqlite dependency

### LINKED ISSUE #237: Sort out SQLite redistribution
We currently redistribute SQLite, but only for UWP and Windows. On OSX and Linux, we assume it is already installed globally.

Ideally, we'd like to get out of the business of redistributing SQLite, and either (1) do nothing (not the best experience on Windows), or (2) depend on an official, SQLite-owned NuGet package.

We tried, with the SQLite team, to produce a NuGet 3 package--[sqlite](https://www.nuget.org/packages/sqlite/3.9.1-test)--for redistributing it, but there was no way do differentiate between a native library that can only run inside of an App container (the WinRT/UWP ones) and the other ones (Win32). (See [this conversation](https://github.com/aspnet/Microsoft.Data.Sqlite/issues/128#issuecomment-152007073).) NuGet 3 also fails to handle Full .NET, AnyCPU apps. (See NuGet/Home#904.) We currently do some fancy MSBuild magic in our package to workaround these. We should work with the NuGet and .NET Core teams to figure out a best practice to use in this situation. Then try again with the SQLite team to produce a package that works with Full .NET, UWP, and .NET Core.

cc @mistachkin, @ericsink, @timheuer

### LINKED ISSUE #245: Incorrect native library imported to UWP apps when Microsoft.Data.Sqlite is a transitive dependency
```
Uwp.App/
     Depends on Uwp.Model
Uwp.Model/
     Depends on Microsoft.Data.Sqlite
```
When building appx packages for uwp.app, sqlite3.dll for win32, not winrt, is included in the build output.

## Workaround

Explicitly add Microsoft.Data.Sqlite as a dependency of Uwp.App

```json
  "dependencies": {
    "Microsoft.NETCore.UniversalWindowsPlatform": "5.1.0",
    "Microsoft.Data.Sqlite": "1.0.0-*"
  }
```

Root cause of https://github.com/aspnet/EntityFramework/issues/5069. Hopefully we can resolve this with a solution for #237

natemcmaster:
This change will stop redistribution of sqlite3.dll within this package. This introduces an external dependency on https://www.nuget.org/packages/sqlite. This package will be updated soon to include native binaries for OSX and Linux, as well as the binaries for Win32 and WinRT.

Fix #237
Fix #245 

cc @divega @bricelam


bricelam:
:shipit:

divega:
:+1: nice to see that we are removing some complexity!

PR #254, opened 2016-05-19
Fix #243 - throw when storing double.NaN

### LINKED ISSUE #243: Storing & Retrieving double.NaN causes NullReferenceException (Sqlite)
_From @teo78 on April 8, 2016 6:44_

I'm testing EFCore RC1 using Sqlite, and encountered a problem when storing/retrieving double.NaN values.

I use TPH inheritance and when storing a double.NaN Sqlite stores this a Null value (see: http://system.data.sqlite.org/index.html/tktview?name=2b823bfe08). Then when loading that row EF Throws a NullReferenceException.

Could this be fixed in EFCore somehow so that Null values would be read in as double.NaN?
If not the only option as I see is to make sure that NaN is never stored which would be nasty solution.

Here is a minimal repro example:
```cs
class Program
    {
        static void Main(string[] args)
        {
            using (var db = new NumberContext())
            {
                db.Database.EnsureDeleted();
                db.Database.EnsureCreated();
                var number = new DoubleNumber() { Value = double.NaN };
                db.DoubleNumbers.Add(number);
                db.SaveChanges();
            }
            using (var db = new NumberContext())
            {
                // Throws NullReference exception on the following line due to NaN double value stored as Null in Sqlite
                var list = db.DoubleNumbers.ToList();
            }
        }
    }
    public class BaseNumber
    {
        public int Id { get; set; }
    }
    public class DoubleNumber : BaseNumber
    {
        public double Value { get; set; }
    }
    public class NumberContext : DbContext
    {
        public DbSet<BaseNumber> BaseNumbers { get; set; }
        public DbSet<DoubleNumber> DoubleNumbers { get; set; }
        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlite($"Filename=Test.db");
        }
    }
```

_Copied from original issue: aspnet/EntityFramework#5011_

natemcmaster:
cc @bricelam @divega


src/Microsoft.Data.Sqlite/SqliteParameter.cs
```diff
@@ -266,6 +266,16 @@ internal bool Bind(Sqlite3StmtHandle stmt)
         private static void BindText(Sqlite3StmtHandle stmt, int index, string value) =>
             NativeMethods.sqlite3_bind_text(stmt, index, value, SQLITE_TRANSIENT);
 
+        private static void BindDouble(Sqlite3StmtHandle stmt, int index, double value)
+        {
+            if (double.IsNaN(value))
```
divega: We need to consider the Infinity values as well, right?

natemcmaster: Actually, SQLite handles infinity just fine. It maps CLR infinity <=> +/- 9e999

divega: Cool!

divega:
LGTM, but I will let @bricelam review.

bricelam:
:shipit:

PR #264, opened 2016-06-03
Remove SqliteEngine.UseWinSqlite3

### LINKED ISSUE #262: Move SqliteEngine.UseWinSqlite3 to Internal for now
Given the issues mentioned in #249, the whole point of using `winsqlite3.dll` is kind of moot. We should remove it for RTM and look at better ways of enabling it in the future.

bricelam:
Resolves #262


natemcmaster:
😱 this is a big breaking change. Although "winsqlite3" has issues on UWP, it works just fine on Windows Server Nano and Win 10 desktop. Before we remove this, we should provide a way for users to still provide a way to use "winsqlite3" in those scenarios. Or, we could move the API to ".Internal" since it isn't fully-baked yet.

divega:
Making it internal doesn't sound bad to me. @bricelam thoughts?

bricelam:
Recalling. Let's move it to an Internal namespace.

PR #267, opened 2016-06-03
Move SqliteEngine to Microsoft.Data.Sqlite.Internal

### LINKED ISSUE #262: Move SqliteEngine.UseWinSqlite3 to Internal for now

bricelam:
Resolves #262


divega:
:shipit: assuming there are no major concerns with this approach

cc @rowanmiller

rowanmiller:
:shipit: from my perspective

natemcmaster:
🐑 :it:

PR #270, opened 2016-06-14
Enable some StyleCop documentation rules

bricelam:


src/Microsoft.Data.Sqlite/SqliteCommand.cs
```diff
@@ -67,6 +67,7 @@ public SqliteCommand(string commandText, SqliteConnection connection, SqliteTran
         /// Gets or sets a value indicating how <see cref="CommandText" /> is interpreted. Only
         /// <see cref="CommandType.Text" /> is supported.
         /// </summary>
+        /// <value>a value indicating how <see cref="CommandText" /> is interpreted.</value>
```
bricelam: Oops, will capitalize.

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.ruleset
```diff
@@ -0,0 +1,145 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Rules for Microsoft.Data.Sqlite" Description="Code analysis rules for Microsoft.Data.Sqlite." ToolsVersion="14.0">
+  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
+    <Rule Id="SA1000" Action="None" />
```
natemcmaster: Never used rulesets....is there a way to ignore all except SA1609?

natemcmaster:
LGTM. Do we plan to enable the same thing on other projects?

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.ruleset
bricelam: Rules have a default action. This disables most that are on by default and enables that one that's off by default. I'll see if there's a way to be explicit about which ones are enabled.

bricelam:
I've proposed it to the engineering team. We'll see what happens. We'll at least enable some of them on EF (I have a branch in progress)

natemcmaster:
Ok. :shipit: for this project.

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.ruleset
bricelam: There is no way to do this. You can make the default action of rules stricter, but you can't relax it.

Issue #278, opened 2016-07-15
sqlite3.dll not found in ASP.NET 4 applications


natemcmaster:
Repro

Create ASP.NET 4 web application
Add Microsoft.Data.Sqlite 1.0.0
Add code that calls `SqliteConnection.Open()`

Error
DllImportException "sqlite3.dll" not found

Reason
Despite the build appropriately places sqlite3.dll in to bin/x86 and bin/x64, ASP.NET sets `AppDomain.CurrentDomain.BaseDirectory` to the project root, but NativeLibraryLoader needs to use "$projectRoot/bin"

VitaliyMF:
I've faced the same issue as @natemcmaster, bug can be reproduced when Microsoft.Data.Sqlite is used from .NET 4.5.1 ASP.NET application. The fix is rather trivial: https://github.com/aspnet/Microsoft.Data.Sqlite/pull/280

PR #280, opened 2016-07-28
Fixed 'sqlite3.dll not found' error for NET451 build when used in ASP.NET apps (#278)

VitaliyMF:
I also faced with issue when Microsoft.Data.Sqlite cannot load sqlite3.dll (.NET 4.5.1 build) when used in ASP.NET application. 

This happens because nuget package 'SQLite.Native' correctly places native dll to x86 / x64 folders, but for ASP.NET apps they are placed in the 'bin' folder, and NativeLibraryLoader.TryLoad don't handle that correctly (I've fixed that).

### DIFF
--- a/src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
+++ b/src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
@@ -50,7 +50,11 @@ namespace Microsoft.Data.Sqlite.Utilities
             }
 
             var dllPath = Path.Combine(applicationBase, GetArchitecture(), dllName);
-
+			
+			if (!File.Exists(dllPath)) {
+				dllPath = Path.Combine(applicationBase, "bin", GetArchitecture(), dllName);
+			}
+			
             if (!File.Exists(dllPath))
             {
                 return false;


natemcmaster:
This approach will solve #278, but I think we can do a more general fix that avoids hard-coding "bin" as a search path. The approach I was going to take is to repeat [this logic](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/d47f274446c28c8ab8c13e3b82353506160ac86b/src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs#L45-L62) using the directory of Assembly.Location in addition to AppDomain.BaseDirectory as a search path.

VitaliyMF:
@natemcmaster  I agree that hardcoding 'bin' is not so good :) but Assembly.Location will not work for ASP.NET apps because it will return path ASP.NET temporary files folder, not to the <app root>/bin folder. In your pull request I see that you use 'System.Web.HttpRuntime.BinDirectory' (wrapped with try-catch block), this is definitely better, but it requires reference to System.Web (this is rather strange dependency for ADO.NET provider). 

Maybe it is better to have just one line of simple code that doesn't require any significant changes?.. Anyway this is all about NET451 build and will not affect NETStandard implementation.

Let me know if I should close my pull request.

natemcmaster:
@VitaliyMF thanks for spending time looking into a solution for #278. Sorry for the overlap in my PR and yours. We must have both started coding at the same time. I'm still waiting for review on PR #282, so don't close yours just yet.

Per your comment about the #282: you've identified well that the trade off to hardcoding "bin" is referencing System.Web. I've been careful about how it's coded so that it will still work even if this ADO.NET provider is used outside of a System.Web project. If we find the reference causes issues, we have other solutions, such as reflecting out System.Web.HttpRuntime instead.

natemcmaster:
@VitaliyMF thanks for the proposed fix. However, I'm going to close this PR as the issue was closed in #282. 

Btw, your feedback was helpful. We avoided adding a reference to System.Web by using AppDomainSetup.PrivateBinPath instead. Thanks!

PR #281, opened 2016-07-29
Parse decimal numbers in exponential form

natemcmaster:
Fix https://github.com/aspnet/EntityFramework/issues/6205

cc @divega


bricelam:
:shipit:

PR #286, opened 2016-08-17
Add documentation

natemcmaster:
Wrote some docs to answer questions I get asked most often.

cc @bricelam @rowanmiller

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,5 @@ launchSettings.json
 *.project.lock.json
 *.nuget.props
 *.nuget.targets
+*.userprefs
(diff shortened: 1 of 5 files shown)


natemcmaster:
Ping

bricelam:
Maybe this should go on the wiki for now...

natemcmaster:
It's already there. Reason for adding here is that wiki = docs graveyard.

rowanmiller:
I doubt this will be any less of a graveyard than the wiki 😄. There isn't much content here, so I would probably just inline it all in the README.

Issue #249, opened 2016-05-16
Problems using winsqlite3.dll in UWP applications


divega:
Current builds of Windows 10 include a `winsqlite3.dll` assembly that should be usable in UWP applications. I tried to do it following the [Local SQLite on UWP walkthrough](https://docs.efproject.net/en/latest/platforms/uwp/getting-started.html) and adding a call to `SqliteEngine.UseWinSqlite3()` early on in the application, but I run into a couple of issues:
1. ~~Native calls through interop such as `sqlite3_open_v2()` fail when targeting winsqlite3.dll instead of sqlite3.dll in an UWP application:~~
   - ~~In debug the application crashes with code `0xc0000409`~~ 
   - ~~In release the APIs just fail~~
   
   ~~I don't know the exact reason an that is going to require further investigation but there seems to be some kind of mismatch in the declaration of the interop call. In fact I tried a couple of tentative workarounds that made the application work:~~
   1. ~~Changing the calling convention from `CallingConvention.Cdecl` to  `CallingConvention.StdCall` or `CallingConvention.Winapi`.~~
   2. ~~Switching from our custom marshalling of UTF8 strings to use `[MarshalAs(UnmanagedType.LPStr)]`. Note that this is not an acceptable workaround because it wouldn't work correctly for international characters, but hopefully `[MarshalAs(UnmanagedType.LPUTF8Str)]` will become an option when https://github.com/dotnet/corefx/issues/7804  gets in for .NET Core 1.1.0.~~
2. The assembly `sqlite3.dll` is deployed with the application and loaded in memory regardless. I verified this with Process Explorer. I even tried removing the assembly from disk but then the application fails to start.

bricelam:
I'm not able to repro the first issue on my machine. I wonder if the second issue is caused by [this line of code](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/Interop/VersionedMethods.cs#L14)...

divega:
Glad that you are not able to repro the first problem in your machine. It may be just a problem in my machine :disappointed:

bricelam:
Looks like having `sqlite3.dll` loaded (and requiring it) is a quirk of .NET Native. I'm not sure what we'll be able to do here...

divega:
Maybe we need to wait to have the NoNative packages? Or is this enough to prioritize the NoNative packages for 1.0.0?

cc @rowanmiller

[...]

Issue #148, opened 2015-10-05
Allow using an external version of sqlite3.dll


rowanmiller:
We ship a version in the NuGet package, but we should provide a way to use a version that is obtained from elsewhere.

Eilon:
There might be some NuGet 3 features that could help with this, such as using Runtime IDs (`rid`s). Should discuss with some NuGet experts because this *might* be super easy.

natemcmaster:
Related: we can start falling back to `C:\Windows\System32\winsqlite3.dll` (which is ships in Win10 installations) if we are unable to load a nuget sqlite3.dll

Eilon:
Is that copy of SQLite meant for public consumption, or is it a private copy for Windows?

natemcmaster:
AFAIK yes. On my machine, it is SQLite 3.8.8.3, x64. It works with all of the tests in this project.

natemcmaster:
(as in yes, meant for public consumption)

Eilon:
Do you know if there's any official-looking statement of this? Or if not, can we contact the owners? I want to make sure if we allow it that there's some level of guarantee around it.

natemcmaster:
I couldn't find anything. I supposed its existence implied it's available for general use. I haven't tested against many versions of Windows, which is why I suggested only as a fallback. The fallback may not be necessary if we can get a good SQLite nuget package.

davidfowl:
I think we should just change this package to use DllImport. The runtime should set the system up in such a way that it will get the right one (architecture wise). For csproj based projects that copy to output would need to preserve the folder structure in the bin folder so that this works. CoreCLR based runtimes can read the project.lock.json and do the right thing.

Assuming that part was figured out, the user could just update to the latest SQLite package package and get the latest bits. DllImport will also search the default search paths so I'm not sure you'd need to do anything else.

bricelam:
@davidfowl This issue is more about the user creating a custom build of SQLite, not just updating to the latest official release.

davidfowl:
Custom build of SQLite.... ok :smile:

bricelam:
lol, yeah... It's a common and supported scenario; see [Compilation Options For SQLite](http://sqlite.org/compile.html) for all the buttons and knobs.

Eilon:
[...]

PR #292, opened 2016-09-28
experimental: switch to SQLitePCL.raw

ericsink:
I consider this more of an FYI than a literal pull request.  

As a proof of concept, I deleted most of the stuff under Interop, added a reference to SQLitePCL.raw, remapped NativeMethods/Sqlite3Handle/Sqlite3StmtHandle, fixed a few things, and it _basically_ works.

I realize that taking a dep on a community open source package like SQLitePCL.raw would be a risk and a big decision.  I'm not pushing for that.  I'm just pleased that it seems to work.

In part, I am looking toward possibilities of an ADO.NET style SQLite API (or maybe even EF Core) on iOS and Android.  Building on SQLitePCL.raw makes that straightforward.  It also provides a path toward things like custom SQLite builds, or swapping the SQLite library for something like SQLCipher.  

In a nutshell, this approach would allow Microsoft.Data.Sqlite to not care about the specific instance of the native SQLite library.  You could just target, say, netstandard1.1 and let SQLitePCL.raw deal with all the drama of Xamarin pinvokes.

My original commit message:

experimental idea.  change Microsoft.Data.Sqlite to build on top of SQLitePCL.raw.  this was mostly done by putting 'using NativeMethods = SQLitePCL.raw;' (and similar for Sqlite3Handle and Sqlite3StmtHandle) at the top of source files.  a few other minor code changes were necessary as well.  plus a couple of new things in SQLitePCL.raw itself, now planned for the 1.1.x release of same.  ability to use winsqlite3.dll removed, as this capability is built-into and handled-by SQLitePCL.raw.  this makes it possible for Microsoft.Data.Sqlite to no longer care about which particular instance of the SQLite library is being used.  most tests pass.  3 test failures.  2 of them look irrelevant, 1 needs further investigation.


bricelam:
Out of curiosity, do you support binding to winsqlite3.dll (ships as part of the Windows SDK)? Does it all work with .NET Native? See #249 for the issues we've ran into.

ericsink:
I haven't tested with .NET Native yet, but I am interested in doing so.  The work I've done to be compatible with other AOT environments makes me optimistic about those aspects, but there could be other problems lurking.

SQLitePCL.raw has a system of "providers", which are implementations of ISQLite3Provider.  Each implementation corresponds to a particular instance of the SQLite native library.  The main assembly needs one of these providers to be injected as part of initialization.

winsqlite3.dll is handled by a different provider.  Same goes for sqlcipher, and for bundled SQLite builds to satisfy Android N, and so on.

ericsink:
Wait.  Does support for UWP10 mean that I actually *have* tested with .NET Native?

src/Microsoft.Data.Sqlite/project.json
```diff
@@ -21,7 +21,7 @@
     ]
   },
   "dependencies": {
-    "SQLite": "3.12.2",
+    "SQLitePCLRaw.bundle_e_sqlite3": "1.1.0-pre20160928150051",
```
bricelam: How would the consumer swap this out for another one?

bricelam: Or would we need to do [something like this](https://github.com/aspnet/Microsoft.Data.Sqlite/issues/148#issuecomment-223383478)?

natemcmaster:
FWIW I think this is an idea we've floated before. https://github.com/aspnet/Microsoft.Data.Sqlite/issues/21. I think this could be a good move. We haven't yet made our library work on AndroidN/iOS. It looks like SQLitePCL.raw has already solved some platforms problems we haven't even discovered.

But as mentioned above, we should check carefully this wouldn't regress our support UWP and .NET Native.

> Wait. Does support for UWP10 mean that I actually have tested with .NET Native?

@ericsink not necessarily. .NET Native tool chain is disabled in "Debug" mode. Did you compile and run in "Release" mode? Did you dry deploying to x86/x64/ARM devices?

test/Microsoft.Data.Sqlite.Tests/TestUtilities/WinSqlite3TestFixture.cs
```diff
@@ -1,27 +0,0 @@
...
-using System.Reflection;
-using Microsoft.Data.Sqlite.Internal;
-using Xunit.Abstractions;
-using Xunit.Sdk;
-
-[assembly: Xunit.TestFramework("Microsoft.Data.Sqlite.WinSqlite3TestFramework", "Microsoft.Data.Sqlite.Tests")]
-
-namespace Microsoft.Data.Sqlite
-{
-    public class WinSqlite3TestFramework : XunitTestFramework
```
ericsink: winsqlite3.dll is supported, but it's done differently.  I didn't set that up as part of the proof-of-concept stuff I was doing, but it's not difficult.

src/Microsoft.Data.Sqlite/project.json
ericsink: There are a couple of options here.

The technique shown in the PR was to take the dependency on a bundle.  This allows me to call SQLitePCL.Batteries_V2.Init() from Microsoft.Data.Sqlite itself (in a static constructor) in portable code.  The bait-and-switch technique replaces this with a real implementation.  This is accomplished by adding the bundle to the tests assembly.

The main reason I used the technique above was that it made the xUnit tests work without sprinkling Init() calls into them.

Another option would be to make Microsoft.Data.Sqlite depend on SQLitePCLRaw.core (the main assembly) and to do the Init() call somewhere else, more explicitly.

ericsink: Ah, yes, the discussion over in item #148 mentions some of the same techniques that I've needed to use.

ericsink:
Embarrassingly enough, I've been thinking of .NET Native as yet another .NET thing that I need to test and figure out.  I've heard inklings of UWP builds being AOT, but I never realized the connection.

I've done some fixes specifically for UWP users who have reported problems, but there have definitely been some holes in my testing along these lines.

I'll report back after I do some additional testing with .NET Native and UWP and release builds and such.

natemcmaster:
> Embarrassingly enough, I've been thinking of .NET Native as yet another .NET thing that I need to test and figure out.

...well, actually...there is CoreRT project (https://github.com/dotnet/corert) which yet another AOT thing...but we don't support it yet. For now, by .NET Native we mean "AOT for Universal Windows Platform".

ericsink:
Thanks for the tip on the .NET Native/UWP/release-mode stuff.  It turns out I did have a bug there.  It showed up only for the winsqlite3 provider.  The e_sqlite3 provider was fine.  Anyway, the compiler error message recommended ExactSpelling=true on the DllImport attributes, and that was the fix.  So my test suite is now passing for UWP release mode builds.

bricelam:
👍 I'm all for making this move too. It offloads most of the platform-specific problems we have and will encounter. We just need to sort out the logistics... I'm marked #21 for re-triage to discuss.

ericsink:
Sounds good.  

Interesting to read issue #21 and see how much has changed in the TWO YEARS since.  :-)

The 1.1.0 release of SQLitePCL.raw should happen next week, and will contain fixes that Microsoft.Data.Sqlite would need.

I will let you guys drive this issue as you see fit.  Just let me know if/how/when I can help.

bricelam:
It looks like the Azure SDK team has taken a dependency on SQLitePCL.raw (in [Microsoft.Azure.Mobile.Client.SQLiteStore](https://www.nuget.org/packages/Microsoft.Azure.Mobile.Client.SQLiteStore)), so we shouldn't run into any problems.

ericsink:
> It looks like the Azure SDK team has taken a dependency on SQLitePCL.raw

Yes, I believe the "Android N SQLite Apocalypse" was what pushed them toward the change.

> so we shouldn't run into any problems

Or rather, if you *do* have any problems, you should be in excellent company.  :-)

rowanmiller:
Closing, but we'll use this as info when we look at moving.

PR #295, opened 2016-10-18
:arrow_up: SQLite to 3.12.3

natemcmaster:
Because of the impending doom of PJ. 3.12.3 includes a fix for https://github.com/natemcmaster/libsqlite3-package/issues/8, which is that 3.12.2 uses the existence of a file named "project.json" to determine if it is in nuget3 or not...which is not longer the right way to detect this.

cc @bricelam


bricelam:
:shipit:

PR #302, opened 2016-11-08
Update partner packages

JunTaoLuo:
Not sure if this package is still needed, but if it is, we need to update its version number.

cc @pranavkm


pranavkm:
:shipit:

PR #314, opened 2017-01-05
Upgrade to VS 2017

natemcmaster:
Removes the Microsoft.Data.Sqlite.FunctionalTest.UWP as this was stale and causing issues with our CI.

TODO: figure out why the 'UseWinSqlite3_throws_when_loaded' test fails on command line. (Will open an issue to follow up on this. Test skipped for now).

cc @divega


natemcmaster:
Ping @bricelam. Can you confirm stylecop is working as you expect?

PR #320, opened 2017-02-10
Use SQLitePCL.raw

### LINKED ISSUE #21: Move to SQLitePCL.raw

### LINKED ISSUE #148: Allow using an external version of sqlite3.dll

### LINKED ISSUE #185: Windows 8 support?
How ready is this to be used? Particularly can it be used in Windows RT apps? 

I tried to install the package at https://www.nuget.org/packages/Microsoft.Data.SQLite/ to my Windows 8 app project and the package failed to be installed, saying that no compatible target platform is found in the package.

### LINKED ISSUE #249: Problems using winsqlite3.dll in UWP applications

### LINKED ISSUE #255: Include Xamarin Android and iOS in NuGet Packages
I tried successfully your excellent new library with Xamarin Andoid. But I need to add a reference manually. Can you configure your NuGet package definition to include **MonoAndroid** and **Xamarin.iOS** (see https://developer.xamarin.com/guides/cross-platform/advanced/nuget).
All dependencies are already included in Mono framework.

See here a sample of the NuGet Package Explorer result.

image

bricelam:
Changes:
* Switches NuGet dependency from SQLite to SQLitePCLRaw.bundle_green
* Splits into two packages:
    * Microsoft.Data.Sqlite--metapackage includes ADO.NET provider and native implementation of SQLite
    * Microsoft.Data.Sqlite.Core--just the ADO.NET provider (resolves #148, fixes #249)
* Removes SqliteEngine
* Updates SqliteConnection and SqliteDataReader Handle property return type from IntPtr to the SQLitePCL.raw primitives
* Stop normalizing relative paths (handled by SQLitePCL.raw)
* Target .NET Standard 1.2 (resolves #185)
* Bump minimum supported SQLite version to 3.7.15
* Removes the use of `ADONET_DATA_DIR` (see aspnet/EntityFramework#7588)
* Other minor cleanups including C# 7 usage

Other notes:
* If a SQLitePCL.raw bundle is installed, Microsoft.Data.Sqlite.Core with auto-initialize it

Resolves #21, resolves #255


src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.csproj
```diff
@@ -1,34 +1,14 @@
...
-Microsoft.Data.Sqlite.SqliteException
-Microsoft.Data.Sqlite.SqliteFactory
-Microsoft.Data.Sqlite.SqliteParameter
-Microsoft.Data.Sqlite.SqliteTransaction</Description>
-    <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <PackageTags>SQLite;Data;ADO.NET</PackageTags>
-    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)$(MSBuildProjectName).ruleset</CodeAnalysisRuleSet>
-    <IncludeSymbols>true</IncludeSymbols>
+    <Description>Metapackage that includes Microsoft.Data.Sqlite.Core and SQLitePCLRaw.bundle_green.</Description>
```
bricelam: cc @rowanmiller

ErikEJ:
Wonder if the newer required runtime version will cause issue on non MS OSes - I recall a number of issues already with the current required version

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.csproj
ErikEJ: SQLite ADO.NET provider, packaged for most common use cases

bricelam:
@ErikEJ It was a problem before we started distributing SQLite on OSX and Linux. You can still hit issues, but only if you step outside of the default bundle_green experience or are deploying to older iOS devices.

Microsoft.Data.Sqlite.sln
```diff
@@ -1,66 +1,43 @@
...
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.Sqlite.Tests", "test\Microsoft.Data.Sqlite.Tests\Microsoft.Data.Sqlite.Tests.csproj", "{69D2DA13-6774-43A9-A2E3-709594136FF8}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.Sqlite", "src\Microsoft.Data.Sqlite\Microsoft.Data.Sqlite.csproj", "{298DB9C7-8C81-40A7-B042-20D62C914820}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
-		Debug|ARM = Debug|ARM
-		Debug|x64 = Debug|x64
```
natemcmaster: 👏 thanks for the cleanup.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
```diff
@@ -0,0 +1,43 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\build\common.props" />
+  <PropertyGroup>
+    <AssemblyName>Microsoft.Data.Sqlite</AssemblyName>
+    <RootNamespace>$(AssemblyName)</RootNamespace>
+    <PackageId>$(MSBuildProjectName)</PackageId>
+    <Description>SQLite implementation of the System.Data.Common provider model.
```
natemcmaster: I know you removed the Summary tag a cleanup commit, but consider re-adding it. NuGet will issue build warnings for long descriptions like this. In general, I'm in favor of eliminating warnings.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
```diff
@@ -0,0 +1,43 @@
...
+    <None Include="stylecop.json" />
+    <None Include="$(CodeAnalysisRuleSet)" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="SQLitePCLRaw.core" Version="1.1.2" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.0.0">
+      <PrivateAssets>All</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="System.Data.Common" Version="4.4.0-*" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
```
natemcmaster: Use 4.4.0-* of this too.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
```diff
@@ -0,0 +1,43 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Update="SqliteCommand.cs">
+      <SubType>Component</SubType>
```
natemcmaster: I think I saw VS automatically add these...but what is it?

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -300,14 +297,14 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
 
                 // NB: This is only a heuristic to separate SELECT statements from INSERT/UPDATE/DELETE statements. It
                 //     will result in unexpected corner cases, but it's the best we can do without re-parsing SQL
-                if (NativeMethods.sqlite3_stmt_readonly(stmt) != 0)
+                if (raw.sqlite3_stmt_readonly(stmt) != 0)
                 {
-                    stmts.Enqueue(Tuple.Create(stmt, rc != SQLITE_DONE));
+                    stmts.Enqueue((stmt, hasRows: rc != raw.SQLITE_DONE));
```
natemcmaster: i ❤️ c# 7.0

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -199,37 +195,18 @@ public override void Open()
...
-            MarshalEx.ThrowExceptionForRC(rc, _db);
+            var rc = raw.sqlite3_open_v2(filename, out _db, flags, vfs: null);
+            SqliteException.ThrowExceptionForRC(rc, _db);
 
             SetState(ConnectionState.Open);
         }
 
-#if !NET451
-        private static string BaseDirectory
-            => Environment.GetEnvironmentVariable("ADONET_DATA_DIR")
```
natemcmaster: Don't we need this for EF migrations to keep working? cref https://github.com/aspnet/EntityFramework.Tools/blob/678ab9a85aa804500c874b175a0000b277de4a12/src/ef/OperationExecutorBase.cs#L14

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -199,37 +195,18 @@ public override void Open()
...
-                ?? ApplicationDataHelper.LocalFolderPath
```
natemcmaster: Does SQLitePCL raw automatically use this for UWP apps?

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -199,37 +195,18 @@ public override void Open()
...
-                ?? AppContext.BaseDirectory;
-#else
-        private static string BaseDirectory
-            => AppDomain.CurrentDomain.GetData("DataDirectory") as string
```
natemcmaster: Likewise: is this something SQLitePCL will use?

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -21,29 +16,26 @@ namespace Microsoft.Data.Sqlite
     /// </summary>
     public class SqliteDataReader : DbDataReader
     {
-        // TODO can be Array.Empty<T>() when upgrading to net46
```
natemcmaster: TODO for netstandard2.0?

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -44,16 +43,13 @@ public SqliteConnection(string connectionString)
...
-            => _db;
-
         /// <summary>
         /// Gets a handle to underlying database connection.
         /// </summary>
         /// <value>A handle to underlying database connection.</value>
         /// <seealso href="http://sqlite.org/c3ref/sqlite3.html">Database Connection Handle</seealso>
-        public virtual IntPtr Handle
-            => _db?.DangerousGetHandle() ?? IntPtr.Zero;
+        public virtual sqlite3 Handle
```
natemcmaster: When API check is re-enabled, this will probably fail here. cref https://github.com/aspnet/KoreBuild/pull/163

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
```diff
@@ -1,72 +0,0 @@
...
-using System.Runtime.InteropServices;
-using System.Text;
-
-using static Microsoft.Data.Sqlite.Interop.Constants;
-
-namespace Microsoft.Data.Sqlite.Interop
-{
-    internal static class MarshalEx
-    {
-        public static string PtrToStringUTF8(IntPtr ptr)
```
natemcmaster: General question: does SQLitePCL default to UTF8 encoding of strings?

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
```diff
@@ -1,279 +0,0 @@
...
-                // prevent re-entrant calls
-                return;
-            }
-            _isLoaded = true;
-            if (ApplicationDataHelper.TemporaryFolderPath != null)
-            {
-                using (var connection = new SqliteConnection("Data Source=:memory:"))
-                {
-                    connection.Open();
-                    connection.ExecuteNonQuery("PRAGMA temp_store_directory = '" + ApplicationDataHelper.TemporaryFolderPath + "';");
```
natemcmaster: This was required for UWP. Handled by SQLitePCL?

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.csproj
natemcmaster: Can we keep the description the same? The fact that it is a metapackage is an implementation detail that 99% of people won't care about.

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.csproj
```diff
@@ -1,34 +1,17 @@
...
+    <TargetFrameworks>net451;netstandard1.2</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
-    <None Include="stylecop.json" />
-    <None Include="$(CodeAnalysisRuleSet)" />
-  </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="SQLite" Version="3.12.3" />
-    <PackageReference Include="StyleCop.Analyzers" Version="1.0.0">
+    <PackageReference Include="SQLitePCLRaw.bundle_green" Version="1.1.2" />
```
natemcmaster: Set `PrivateAssets=None`. SQLitePCLRaw.bundle_green has some build targets that should be included. Cref https://github.com/NuGet/Home/issues/4521

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
```diff
@@ -1,149 +0,0 @@
...
-            yield return AppDomain.CurrentDomain.BaseDirectory;
-            yield return Path.Combine(AppDomain.CurrentDomain.BaseDirectory, GetArchitecture());
-
-            var privateBinPath = AppDomain.CurrentDomain.SetupInformation.PrivateBinPath;
-            if (!string.IsNullOrEmpty(privateBinPath))
-            {
-                foreach (var path in privateBinPath.Split(new [] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries))
-                {
-                    yield return path;
-                    yield return Path.Combine(path, GetArchitecture());
```
natemcmaster: How does SQLitePCL handle AnyCPU desktop apps?

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
Eilon: It makes a different icon show up in Solution Explorer 😄

Eilon:
@natemcmaster re: new 3rd party dependency. Yup, already noted and we filed the paperwork with the paperwork people 😄

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.csproj
Eilon: In my view the term "metapackage" is a bit nerdy so I agree that's somewhat of an implementation detail. The most interesting things to put in the description are: 1. What is this thing? 2. Why do I need/want it?

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
natemcmaster: Nvm. Just realized you already opened https://github.com/ericsink/SQLitePCL.raw/issues/139. In the meantime, let's ensure we don't break this.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: Yes

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: I need to add this one back.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: Filed aspnet/EntityFramework#7588 to sort this out. Ideally, I just want to rely on the current directory.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: Yep. Will update baselines

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
bricelam: Only if we decide to drop support for older platforms which I'm not sure we should with this library...

bricelam: Personally, I want to position this as just a .NET Standard 1.2 library.

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
bricelam: Yep

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
bricelam: Yep.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
ericsink: It has some code to try and resolve things at runtime.  In sqlite3_pinvoke.cs, look for PRELOAD_FROM_ARCH_DIRECTORY.  I can't claim that code is perfect, but it is generally working for people.

src/Microsoft.Data.Sqlite/Interop/NativeMethods.cs
ericsink: Yes.

src/Microsoft.Data.Sqlite/Interop/MarshalEx.cs
ericsink: Yes.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
natemcmaster: Until we have the fix in EF tooling, can we keep this in?

I highly-suspect that this was only ever used by EF tooling, but we did document it for 1.x. It's possible this will be a breaking change for someone out there. https://github.com/aspnet/Microsoft.Data.Sqlite/wiki/Connection-Strings

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
bricelam: Looks like the same way we did: `x86` and `x64` directories + `LoadLibraryEx`.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
bricelam: Nothing was handling `<Summary>`. NuGet [doesn't support it](https://docs.microsoft.com/en-us/nuget/schema/msbuild-targets)... (I was getting the warning even with it.)

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
bricelam: Does it exist? I was having problems when I tried...

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
bricelam: It's because they inherit from `Component` on `net451`. I figured I'd stop fighting it and just check it in.

bricelam: It's also a hint that there may be a designer associated with them.

src/Microsoft.Data.Sqlite/Microsoft.Data.Sqlite.csproj
bricelam: Will keep the same. I debated doing that, but wanted emphasize that one includes SQLite while the other doesn't, but it probably doesn't actually matter.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
natemcmaster: I'm trying to navigate the layers of SQLitePCL packages. I'm sure this handled, but can you point me to which package actually has the native windows x64 vs x86 binaries? What I'm trying to make sure we've addressed this this: https://github.com/dotnet/sdk/pull/847. Microsoft.NET.Sdk will not require .NET Framework apps to specify a runtime identifier. But if the package is using the 'runtimes' folder of NuGet to distribute win-x86 vs win-x64 binaries, we may run into issues.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
natemcmaster: That's annoying. Consider opening and issue on dotnet/roslyn-project-system.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: lol, this feature is a bit hard to grok syntactically.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
ericsink: I think you're looking for SQLitePCLRaw.lib.e_sqlite3.v110_xp

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
natemcmaster: Oh, nvm. Weird...was sure NuGet supported that property.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: Yes, will keep for now.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
natemcmaster: Yes, it's on the https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.ValueTuple. I updated our mirror settings to include this package. I manually added it to our feeds to unblock this.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: Actually, without this, it's functionally equivalent. It will use the current directory which is the same as the data directory.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
natemcmaster: Have you tried this package with VS 2017 and the .NET Core SDK? I suspect you are going to run into this issue: https://github.com/NuGet/Home/issues/4432. Basically, because the targets file adds the library as 'Content', it will end up in nuget packages. You can probably make a simple change to fix this. Set `<Pack>false</Pack>` on the Content items. You may also want to set CopyToPublishDirectory in addition to output directory.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
natemcmaster: Regardless, I'm in favor of keeping it. It's a useful internal trick to have up our sleeves. I see no harm in leaving it, even if we stop using it.

src/Microsoft.Data.Sqlite/Utilities/NativeLibraryLoader.cs
ericsink: No, I have not yet tested with VS 2017.  Until a couple of days ago when the March release date was announced, I thought I had plenty of time.  :-)

Thanks for the tip.  I will investigate.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
bricelam: Awesome, thank you. Tuple power! ✊

ericsink:
FYI about bundle_green:

I currently have three bundles:

1.  bundle_e_sqlite3 (always uses my e_sqlite3 builds)
2.  bundle_sqlcipher (uses sqlcipher builds, which I currently get from couchbaselabs/couchbase-lite-libsqlcipher )
3.  bundle_green

Basically, bundle_green means "Use the e_sqlite3 builds everywhere except iOS.  On iOS, use the SQLite provided by the system."

In other words, it is identical to bundle_e_sqlite3, except on iOS.

Why is it called "green"?  Because it provides the behavior that Frank Krueger aka @praeclarum wanted for sqlite-net-pcl, and it didn't seem right to call it `bundle_frank`, and `bundle_sqlite_net` is confusing, and `bundle_use_e_sqlite3_everywhere_except_ios` is too long, and Frank is from the pacific northwest where there are lots of trees, and trees are mostly green, and I had already spent too much time deciding on the name, so I just settled it and moved on.

Anyway, people like bundle_green, except one issue tends to come up.  My e_sqlite3 builds are compiled with SQLite's default setting for thread safety, but the iOS system SQLite is compiled with a different setting.  The main difference is that the iOS setting means that a sqlite3 connection handle cannot be shared across threads.  Every so often, somebody has trouble because their app is giving different behavior on iOS than on other platforms.

Luckily, the thread safety setting can be changed at runtime, using `sqlite3_config()`.  So, people can eliminate the inconsistency by changing the setting to one way or the other.  However, `sqlite3_config()` needs to be called before any other sqlite function, and this is why those `sqlite3_win32_set_directory` calls in sqlite3_invoke.cs (see related comments elsewhere on this PR) are problematic.  Currently, with SQLitePCL.raw on UWP, to call `sqlite3_config`, you have to (carefully) do a `sqlite3_shutdown` and then call `sqlite3_config` and then call `sqlite3_initialize`.  Kinda messy, but it works, and only UWP has this problem.

Anyway, I tend to use bundle_e_sqlite3 in my own apps.  Having consistent compile settings for the SQLite library seems more important to me than making the size of the iOS app smaller by using the system SQLite on that platform.

Anyway, if you stay with bundle_green, you may want to add the necessary `sqlite3_config()` calls to keep the thread safety behavior consistent across all platforms.  OTOH, if you never share a sqlite3 connection handle across threads, those config calls probably are not necessary.

Switching to bundle_e_sqlite3 is an option as well.

Finally, it is worth noting that I would be happy to add (and maintain) another bundle which is customized for your requirements.  Suppose you decide you want the same compile options on all platforms (like e_sqlite3), but you want compile options that are different from e_sqlite3.  For example, I build e_sqlite3 with support for FTS, RTREE, and json.  The builds would be smaller if any of those were removed.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: lol, I made a best effort to add them. Will update if they break.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: The "harm" is that I'd have to target netstandard1.3 and #185 would remain unresolved. I think I'll take it out for now, and we can see if anyone complains.

bricelam:
@ericsink TBH, the only reason I picked `SQLitePCLRaw.bundle_green` is because [Microsoft.Azure.Mobile.Client.SQLiteStore](https://www.nuget.org/packages/Microsoft.Azure.Mobile.Client.SQLiteStore) uses it, and I didn't want to create an ecosystem where mixing different SQLite technologies in the same project created NuGet package hell. 😈

ericsink:
@bricelam Makes sense.  I believe I talked with them about the tradeoffs as well.  I just want folks to be informed.

natemcmaster:
@ericsink what is your release timeline like for updates? I'm hesitant to sign off on this change now if we can't get a fix for https://github.com/ericsink/SQLitePCL.raw/issues/140#issuecomment-279474411 before our next release.

The issue means anyone with VS 2017 and a class library referencing Microsoft.Data.Sqlite will end up with sqlite3.dll in their NuGet package when they call "dotnet pack".

ericsink:
@natemcmaster I've been traveling, but this will be a priority when I get back in the office (tomorrow-ish). I would expect this to be days, or worst case, weeks.

When is your next release?  I've had the impression it was still months away.  Your concern makes me think I am misunderstanding something.

natemcmaster:
> days, or worst case, weeks.

That's what I was hoping for, but wanted to double check. In that case, we will be fine. Our next release will be after that.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
natemcmaster: FYI I've opened https://github.com/NuGet/Home/issues/4587 as this long description will always produce a build warning.

Issue #235, opened 2016-02-28
Implement SqliteCommand.Prepare


buybackoff:
Prepare method is [not implemented](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/SqliteCommand.cs#L85) and the prepared statement is not reused ([as in System.Data.SQLite](https://github.com/OpenDataSpace/System.Data.SQLite/blob/master/System.Data.SQLite/SQLiteCommand.cs#L892)), but [is disposed](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/SqliteCommand.cs#L205) on every execution. Currently there is no way to reuse a command like in [this example](https://msdn.microsoft.com/en-us/library/system.data.idbcommand.prepare%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396).

In the simplest [insert case](https://github.com/Spreads/Spreads/blob/master/tests/Spreads.Extensions.Tests/Storage/SQLite/SqlitePerformanceTest.cs#L39), it takes almost 20% (of which UTF8 marshalling is 2.4 pp):

![image](https://cloud.githubusercontent.com/assets/489852/13377623/4ba8919e-ddf3-11e5-9fba-298bb878fba8.png)

I haven't found any discussion in issues about that. It is not hard to implement, but I am not sure if it is already on your TODO list and what design you have in your minds?

buybackoff:
Why not using finalizers for this? Given that SQLite does not support stored procedures, prepared statements functionality is the only opportunity and performance cost is quite substantial.

bricelam:
@buybackoff Oh I wasn't pushing back on implementing the feature in some other way, I was just giving some history.

buybackoff:
@bricelam so will you add this back some day?

rowanmiller:
Moving to the backlog as we do want to keep this around to consider implementing in the future

roji:
@bricelam regarding your comment on disposal and memory looks above...

Isn't the idea of preparing of a prepared command (almost) always imply, by definition, some external resource being allocated which later has to be freed? In the case of a remote database server like PostgreSQL, resources on the server are allocated when a command is prepared (that's what makes the speedup possible), in sqlite it's in unmanaged memory in the same process, but it's always going to be there...

[...]

PR #334, opened 2017-03-17
Implement SqliteCommand.Prepare.

AlexanderTaeschner:
Implements SqliteCommand.Prepare according to the remarks in the issue.

Addresses #235

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
+++ b/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
@@ -39,6 +39,7 @@ Microsoft.Data.Sqlite.SqliteTransaction</Description>
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.Data.Common" Version="$(CoreFxVersion)" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
+++ b/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
@@ -39,7 +39,7 @@ Microsoft.Data.Sqlite.SqliteTransaction</Description>
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.Data.Common" Version="$(CoreFxVersion)" />
-    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
+    <PackageReference Include="System.ValueTuple" Version="$(CoreFxVersion)" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
@@ -215,12 +215,12 @@ namespace Microsoft.Data.Sqlite
             if (Connection == null
                 || Connection.State != ConnectionState.Open)
             {
-                throw new InvalidOperationException(Resources.CallRequiresOpenConnection("Prepare"));
+                throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(Prepare)));
             }
 
             if (string.IsNullOrEmpty(CommandText))
             {
-                throw new InvalidOperationException(Resources.CallRequiresSetCommandText("Prepare"));
+                throw new InvalidOperationException(Resources.CallRequiresSetCommandText(nameof(Prepare)));
             }
 
             if (_preparedStatements.Count == 0)
--- a/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
@@ -21,6 +21,8 @@ namespace Microsoft.Data.Sqlite
     {
         private readonly Lazy<SqliteParameterCollection> _parameters = new Lazy<SqliteParameterCollection>(
             () => new SqliteParameterCollection());
+        private readonly List<sqlite3_stmt> _preparedStatements = new List<sqlite3_stmt>();
+        private string _commandText;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SqliteCommand" /> class.
@@ -82,7 +84,18 @@ namespace Microsoft.Data.Sqlite
         /// Gets or sets the SQL to execute against the database.
         /// </summary>
         /// <value>The SQL to execute against the database.</value>
-        public override string CommandText { get; set; }
+        public override string CommandText
+        {
+            get { return _commandText; }
+            set
+            {
+                if (!value.Equals(_commandText))
+                {
+                    Connection?.DisposePreparedStatements(_preparedStatements);
+                    _commandText = value;
+                }
+            }
+        }
 
         /// <summary>
         /// Gets or sets the connection used by the command.
@@ -152,6 +165,22 @@ namespace Microsoft.Data.Sqlite
         public override UpdateRowSource UpdatedRowSource { get; set; }
 
         /// <summary>
+        /// Releases any resources used by the connection and closes it.
+        /// </summary>
+        /// <param name="disposing">
+        /// true to release managed and unmanaged resources; false to release only unmanaged resources.
+        /// </param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                Connection?.DisposePreparedStatements(_preparedStatements);
+            }
+
+            base.Dispose(disposing);
+        }
+
+        /// <summary>
         /// Creates a new parameter.
         /// </summary>
         /// <returns>The new parameter.</returns>
@@ -170,6 +199,21 @@ namespace Microsoft.Data.Sqlite
         /// </summary>
         public override void Prepare()
         {
+            if (Connection == null
+                || Connection.State != ConnectionState.Open)
+            {
+                throw new InvalidOperationException(Strings.CallRequiresOpenConnection("Prepare"));
+            }
+
+            if (string.IsNullOrEmpty(CommandText))
+            {
+                throw new InvalidOperationException(Strings.CallRequiresSetCommandText("Prepare"));
+            }
+
+            if (_preparedStatements.Count == 0)
+            {
+                Connection.GetPreparedStatements(CommandText, _preparedStatements);
+            }
         }
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


bricelam:
P.S. I love the decision to make `Prepare` throw if all the statements can't be eagerly compiled, but still lazily compile them during `Execute*`

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -325,5 +328,61 @@ public virtual void EnableExtensions(bool enable = true)
...
+                // Statement was empty, white space, or a comment
+                if (stmt.ptr == IntPtr.Zero)
+                {
+                    if (!string.IsNullOrEmpty(tail))
+                    {
+                        continue;
+                    }
+
+                    break;
+                }
```
bricelam: Some of this code is duplicated in `ExecuteReader`. Can we share more of the implementation somehow?

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -151,6 +164,22 @@ protected override DbParameterCollection DbParameterCollection
...
+        /// </summary>
+        /// <param name="disposing">
+        /// true to release managed and unmanaged resources; false to release only unmanaged resources.
+        /// </param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                Connection?.DisposePreparedStatements(_preparedStatements);
+            }
```
bricelam: If the connection is open and this object gets garbage collected, will the statements be freed or do they remain until the connection is closed?

bricelam: Instead of letting the connection own the statements, we may need to keep a weak reference to the commands and Dispose of any lingering ones up when the connection is closed.

bricelam:
Another scenario to test:
* Execute works when command text includes dependent commands

bricelam:
We could optimize `ExecuteReader` even more to not re-bind the parameters every time. They only need to be re-bound to the statement when their values change.

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
```diff
@@ -64,9 +68,49 @@ public void CreateParameter_works()
...
+                connection.Open();
+
+                var ex = Assert.Throws<InvalidOperationException>(() => connection.CreateCommand().Prepare());
+
+                Assert.Equal(Strings.CallRequiresSetCommandText("Prepare"), ex.Message);
+            }
+        }
+
+        [Fact]
+        public void Prepare_throws_when_command_text_contains_dependent_commands()
```
bricelam: Should also explicilty test that `Execute` works with dependent statements

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
AlexanderTaeschner: Both the connection and the commands contain lists of the prepared statements. Calling Dispose(true) for the command frees the prepared statements of the single command and calling Close or Dispose of the connection will free all remaining statements prepared for this connection (if dispose for the commands was not called before). The new commit disposes the already prepared statements inside the command when the connection propert of the command is changed.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: I do not see how, since the lengthy part of this duplication is controlling the flow of the while loop. In ExecuteReader this part delivers the single statements used for non prepared execution but steers the loop which is shared by the code for prepared statements. The only possibility would be to leave the empty statements in the list of prepared statements and do the skipping every time inside ExecuteReader.

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
AlexanderTaeschner: Test added in newest commit.

bricelam:
Here is a test illustrating the garbage collection issue I mentioned.
```C#
[Fact]
public void Stmts_are_freed_when_DbCommand_is_finalized()
{
    {
        connection.Open();

        CreateAndPrepareCommand(connection);
        for (int i = 0; i < 500; i++)
        {
            GC.Collect();
        }

        Assert.Empty(connection._preparedStatements);
    }
}

[MethodImpl(MethodImplOptions.NoInlining)]
void CreateAndPrepareCommand(SqliteConnection connection)
{
    command.CommandText = "SELECT 1;";
    command.Prepare();
}
```

AlexanderTaeschner:
Okay I now added a finalizer to the SqliteCommand to solve the garbage collection issue. The SqliteDataReader now gets initialized with the calling instance of the SqliteCommand object in order to prevent the finalizer to be run while the prepared statements are used inside the SqliteDataReader.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -62,6 +65,14 @@ public SqliteCommand(string commandText, SqliteConnection connection, SqliteTran
         }
 
         /// <summary>
+        /// Finalizes an instance of the <see cref="SqliteCommand"/> class.
+        /// </summary>
+        ~SqliteCommand()
+        {
+            Connection?.DisposePreparedStatements(_preparedStatements);
+        }
```
bricelam: You shouldn't need a finalizer. The base implementation calls `Dispose(false)`. Just remove the `if` block in the override below.

AlexanderTaeschner:
In the newest version I use a `List<WeakReference<SqliteCommand>>` in `SqliteConnection` as proposed. Strangely one test fails in AppVeyor, but not on my local machine.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -259,7 +269,11 @@ protected override void Dispose(bool disposing)
         /// transaction.
         /// </remarks>
         public new virtual SqliteCommand CreateCommand()
-            => new SqliteCommand { Connection = this, Transaction = Transaction };
+        {
+            var command = new SqliteCommand { Connection = this, Transaction = Transaction };
+            _commands.Add(new WeakReference<SqliteCommand>(command));
```
bricelam: This list needs to be maintained by the `SqliteCommand.Connection` property. Imagine the following.

```c#
command.Connection = new SqliteConnection("...");
```

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
```diff
@@ -39,6 +39,7 @@ Microsoft.Data.Sqlite.SqliteTransaction</Description>
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.Data.Common" Version="$(CoreFxVersion)" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
```
bricelam: Use `$(CoreFxVersion)` instead of the hard-coded version here

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -229,6 +231,14 @@ public override void Close()
             }
 
             Transaction?.Dispose();
+            foreach (var reference in _commands)
+            {
+                if (reference.TryGetTarget(out var command))
+                {
+                    command.Dispose();
+                }
+            }
```
bricelam: `_commands` should be cleared here.

bricelam:
P.S. I'm very excited about this PR. I think I'll write a blog post about it after writing some micro-benchmarks to quantify its impact.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -295,7 +379,6 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
                 }
                 catch
                 {
-                    stmt.Dispose();
                     throw;
                 }
```
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -82,13 +85,36 @@ public override CommandType CommandType
...
+        public new virtual SqliteConnection Connection
+        {
+            get { return _connection; }
+            set
+            {
+                if (!value.Equals(_connection))
+                {
+                    DisposePreparedStatements();
+                    _connection = value;
+                    _connection.Commands.Add(new WeakReference<SqliteCommand>(this));
```
Z00L: Is there any precondition that says "`value` will never be `null`"? If not, this will cause a NRE.
I mean it could be a use case to erase the command's reference to the connection, dispose the previous/obsolete connection, but keep the command alive to reuse it with an different connection later.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -111,6 +111,7 @@ public new virtual SqliteConnection Connection
                 {
                     DisposePreparedStatements();
                     _connection = value;
+                    _connection.Commands.Add(new WeakReference<SqliteCommand>(this));
```
bricelam: Need to remove it from the old connection (if any). Otherwise, this might get disposed when it shouldn't be.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: I had the same comment. Also need to remove it from the previous connection if it's not `null`.

AlexanderTaeschner: Thanks, very good catch.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -82,13 +85,36 @@ public override CommandType CommandType
...
-        public new virtual SqliteConnection Connection { get; set; }
+        public new virtual SqliteConnection Connection
+        {
+            get { return _connection; }
+            set
+            {
+                if (value != _connection)
+                {
+                    DisposePreparedStatements();
+                    _connection = value;
```
bricelam: Also need to remove it from the previous connection if not null. Otherwise, this command will get disposed when it shouldn't.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -170,6 +209,40 @@ protected override DbParameter CreateDbParameter()
...
         {
+            if (Connection == null
+                || Connection.State != ConnectionState.Open)
+            {
+                throw new InvalidOperationException(Resources.CallRequiresOpenConnection("Prepare"));
+            }
+
+            if (string.IsNullOrEmpty(CommandText))
+            {
+                throw new InvalidOperationException(Resources.CallRequiresSetCommandText("Prepare"));
```
Z00L: Make use `nameof(Prepare)` ? (The same some lines above at line 215)

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
AlexanderTaeschner: In order to find it in the old collection I now added a new field with the weak reference to the SqliteCommand. I'm not sure if this is the best solution.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: Seems OK like this. I may fiddle with the internal APIs a little bit before merging.

bricelam:
FYI, I've made a few cleanups in https://github.com/bricelam/Microsoft.Data.Sqlite/commit/a9ee1d30247c9e0a7898b058dbd622a44f09558b. I was able to remove the duplication between `Prepare` and `ExecuteReader`

bricelam:
Doh! It looks like we caused some downstream failures: https://github.com/aspnet/EntityFramework/issues/7982

Issue #318, opened 2017-01-31
Support reading DateTime values from REAL columns


bricelam:
We should support reading `DateTime` (and `DateTimeOffset`) values from `REAL` columns (in Julian day format).

PR #335, opened 2017-03-21
Add conversion from julian date to SqliteDataReader.GetDateTime

AlexanderTaeschner:
Add conversion from julian date to SqliteDataReader.GetDateTime.

Addresses #318


AlexanderTaeschner:
In #318 reading DateTimeOffset is also mentioned. This is not included here, since it is not possible to store the DateTime and the Offset in a single REAL value. The user has to find a way to store both informations in different columns.

bricelam:
I think it would be OK to support `DateTimeOffset` using `TimeZoneInfo.Local.BaseUtcOffset`. This currently what happens with `TEXT` columns that don't specify an offset.

bricelam:
In other words, make it behave like `new DateTimeOffset(DateTimeConversion.FromJulianDate(value))`

src/Microsoft.Data.Sqlite.Core/Utilities/DateTimeConversion.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    internal static class DateTimeConversion
+    {
+        // Computes DateTime from julian date. This function is a port of the
+        // computeYMD and computeHMS functions from the original Sqlite core
+        // source code in 'date.c'.
+        public static DateTime FromJulianDate(double julianDate)
+        {
+            // computeYMD
+            long iJD = (long)(julianDate * 86400000.0 + 0.5);
```
bricelam: Our [engineering guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#usage-of-the-var-keyword) encourage the use of `var`.

src/Microsoft.Data.Sqlite.Core/Utilities/DateTimeConversion.cs
```diff
@@ -0,0 +1,45 @@
...
+            double fracSecond = s / 1000.0;
+            s = (int)fracSecond;
+            fracSecond -= s;
+            int hour = s / 3600;
+            s -= hour * 3600;
+            int minute = s / 60;
+            fracSecond += s - minute * 60;
+
+            int second = (int)fracSecond;
+            int millisecond = (int)Math.Round((fracSecond - second) * 1000.0);
```
bricelam: Probably overkill to use `Math.Round` here. I'd just let it truncate.

src/Microsoft.Data.Sqlite.Core/Utilities/DateTimeConversion.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Microsoft.Data.Sqlite
+{
+    using System;
+
+    internal static class DateTimeConversion
+    {
+        // Computes DateTime from julian date. This function is a port of the
+        // computeYMD and computeHMS functions from the original Sqlite core
+        // source code in 'date.c'.
+        public static DateTime FromJulianDate(double julianDate)
```
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/Utilities/DateTimeConversion.cs
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/Utilities/DateTimeConversion.cs
AlexanderTaeschner: Using only truncate would result in test failure in the test `GetDateTime_works_with_real`: It would result in 119 (untruncated value is something like 119.999999) instead of 120 milliseconds. If desired I can change it to truncate and adjust the expected test result.

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -384,7 +384,18 @@ public override char GetChar(int ordinal)
...
         /// <returns>The value of the column.</returns>
         public override DateTime GetDateTime(int ordinal)
-            => DateTime.Parse(GetString(ordinal), CultureInfo.InvariantCulture);
+        {
+            var sqliteType = GetSqliteType(ordinal);
+            switch (sqliteType)
+            {
+                case raw.SQLITE_TEXT:
+                    return DateTime.Parse(GetString(ordinal), CultureInfo.InvariantCulture);
+                case raw.SQLITE_FLOAT:
```
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -384,7 +384,18 @@ public override char GetChar(int ordinal)
...
+                    return DateTimeConversion.FromJulianDate(GetDouble(ordinal));
+                default:
+                    throw new InvalidCastException();
```
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/Utilities/DateTimeConversion.cs
bricelam: Ah, let's leave it in then. I forgot how weird native `double`s can be.

Issue #191, opened 2015-12-08
Support reading Guid values from TEXT columns


rowanmiller:
Copied from https://github.com/aspnet/EntityFramework/issues/3950

Currently we assume GUID values are stored in BLOB columns, so you get an exception when we try to parse the value when it is stored as a STRING.

RonFrick:
Thanks Rowan! Is this something that I should jump on with a pull request or will you be able to knock it out fairly quickly?

rowanmiller:
Pull request would be great - we are swamped at the moment.

RonFrick:
First pull request created. :) Although I think I might have picked the wrong stream to pull from, can we change it on the fly before I start working?

ErikEJ:
Think you must start over. First fork, clone, branch

bricelam:
My thoughts on this: If the bytes returned by `GetBlob` are larger than expected, convert them to text and try creating a `Guid` from them.

RonFrick:
Anyone had time to look at this?

RonFrick:
I have it working. I took your suggestion Brice and converted it to text to make sure the length was correct. Not sure if I did this correctly but I checked in my changes here. 

https://github.com/aspnet/Microsoft.Data.Sqlite/compare/dev...RonFrick:ron?expand=1

ErikEJ:
I think you should call GetBlob first, like @bricelam suggests.

RonFrick:
Here? public override Guid GetGuid(int ordinal) => new Guid(GetGuidBlob(ordinal));

This throws an exception…

public override Guid GetGuid(int ordinal) => new Guid(GetBlob(ordinal));

RonFrick:
Is it just the naming that is a problem? We could still call it GetBlob and use an overload?

public override Guid GetGuid(int ordinal) => new Guid(GetBlob(ordinal,true));

private string GetBlob(int ordinal,bool isGuid)
        {
            if (IsDBNull(ordinal))
            {
                throw new InvalidCastException();
            }

            string blob = System.Text.Encoding.UTF8.GetString(NativeMethods.sqlite3_column_blob(_stmt, ordinal));
            if (isGuid && blob.Length > 36)
                return blob.Substring(0, 36);
            else
                return blob;
        }

RonFrick:
This also works but it is not safe?

 public override Guid GetGuid(int ordinal) => new Guid(GetBlob(ordinal).ToString().Substring(0,36));

[...]

PR #339, opened 2017-03-23
Support reading Guid values from TEXT columns

AlexanderTaeschner:
Addresses #191

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
@@ -446,7 +446,26 @@ namespace Microsoft.Data.Sqlite
         /// <param name="ordinal">The zero-based column ordinal.</param>
         /// <returns>The value of the column.</returns>
         public override Guid GetGuid(int ordinal)
-            => new Guid(GetBlob(ordinal));
+        {
+            var sqliteType = GetSqliteType(ordinal);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -446,7 +446,26 @@ public override float GetFloat(int ordinal)
...
+                    if (bytes.Length == 16)
+                    {
+                        return new Guid(bytes);
+                    }
+                    else
+                    {
+                        return new Guid(System.Text.Encoding.UTF8.GetString(bytes, 0, bytes.Length));
+                    }
+                default:
+                    return new Guid(GetBlob(ordinal));
```
bricelam: Hmm... maybe just default to `GetString` if it's not a `BLOB`? Then again, neither exception is really useful...
* *Guid should contain 32 digits with 4 dashes (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx).*
* *Byte array for GUID must be exactly 16 bytes long.*

**But**, the `string` overload *could* be if it told you the value. E.g. "The value '3.14' is invalid. Guid should contain..."

AlexanderTaeschner: Defaulting to GetString seems indeed to be cleaner. Done in the newest commit.

Issue #19, opened 2014-04-02
Custom collations


bricelam:
SQLite allows you to [define new collating sequences](http://www.sqlite.org/c3ref/create_collation.html). We want to enable this by allowing calls like the following.

```C#
connection.CreateCollation(
    "NOCASE",
    (s1, s2) => string.Compare(s1, s2, StringComparison.OrdinalIgnoreCase));

connection.CreateCollation(
    "NOCASE",
    logger, // User state. Avoids capturing in the lambda.
    (l, s1, s2) =>
    {
        l.Log("Collating...");

        return string.Compare(s1, s2, StringComparison.OrdinalIgnoreCase);
    });
```

The definitions for the new API would be:
```C#
class SqliteConnection
{
    public virtual void CreateCollation(
        string name,
        Comparison<string> comparison);

    public virtual void CreateCollation<T>(
        string name,
        T state,
        Func<T, string, string, int> comparison);
}
```

lalibi:
I have a problem with statements like `SELECT * FROM User WHERE UPPER(Title) LIKE '%ΝΙΚΟΣ%'`. To my understanding, that's because functions like `UPPER` and `LOWER` don't support non-ASCII characters.

> `SELECT UPPER('Νίκος')` gives `Νίκος` and not `ΝΙΚΟΣ`  

According to my readings this can be solved with a Custom Collation. I see that this issue was open almost 3 years ago, so I suppose it won't be done anytime soon.

In the meantime, do you know of any workarounds? I found several solutions but nothing I can apply to this library for .NET Core.

bricelam:
In version 2.0, we're moving to [SQLitePCL.raw](https://github.com/ericsink/SQLitePCL.raw) which makes it a lot easier to do.

```C#
using (var connection = new SqliteConnection(connectionString))
{
    connection.Open();
    
    SQLitePCL.raw.sqlite3_create_collation(
        connection.Handle,
        "MY_NOCASE",
        null,
        (_, s1, s2) => string.Compare(s1, s2, StringComparison.OrdinalIgnoreCase));
    
    // TODO: "SELECT 'Νίκος' = 'ΝΙΚΟΣ' COLLATE MY_NOCASE;"
}
```

You can also do this with version 1.x, but you need to define your own P/Invoke of `sqlite3_create_collation`.

AlexanderTaeschner:
[...]

PR #340, opened 2017-03-23
Add SqliteConnection.CreateCollation method

AlexanderTaeschner:
Addresses #19

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
@@ -270,7 +270,7 @@ namespace Microsoft.Data.Sqlite
         /// <param name="name">Name of the collation.</param>
         /// <param name="comparison">Method that compares two strings.</param>
         public virtual void CreateCollation(string name, Comparison<string> comparison)
-            => CreateCollation(name, null, (_, s1, s2) => comparison(s1, s2));
+            => CreateSqlite3Collation(name, null, (_, s1, s2) => comparison(s1, s2));
 
         /// <summary>
         /// Create custom collation.
@@ -280,9 +280,9 @@ namespace Microsoft.Data.Sqlite
         /// <param name="state">State object passed to each invokation of the collation.</param>
         /// <param name="comparison">Method that compares two strings, using additional state.</param>
         public virtual void CreateCollation<T>(string name, T state, Func<T, string, string, int> comparison)
-            => CreateCollation(name, state, (v, s1, s2) => comparison((T)v, s1, s2));
+            => CreateSqlite3Collation(name, state, (v, s1, s2) => comparison((T)v, s1, s2));
 
-        private void CreateCollation(string name, object state, delegate_collation collation)
+        private void CreateSqlite3Collation(string name, object state, delegate_collation collation)
         {
             var rc = raw.sqlite3_create_collation(_db, name, state, collation);
             SqliteException.ThrowExceptionForRC(rc, _db);
--- a/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
@@ -270,7 +270,7 @@ namespace Microsoft.Data.Sqlite
         /// <param name="name">Name of the collation.</param>
         /// <param name="comparison">Method that compares two strings.</param>
         public virtual void CreateCollation(string name, Comparison<string> comparison)
-            => CreateSqlite3Collation(name, null, (_, s1, s2) => comparison(s1, s2));
+            => CreateCollation<object>(name, null, (_, s1, s2) => comparison(s1, s2));
 
         /// <summary>
         /// Create custom collation.
@@ -280,11 +280,8 @@ namespace Microsoft.Data.Sqlite
         /// <param name="state">State object passed to each invokation of the collation.</param>
         /// <param name="comparison">Method that compares two strings, using additional state.</param>
         public virtual void CreateCollation<T>(string name, T state, Func<T, string, string, int> comparison)
-            => CreateSqlite3Collation(name, state, (v, s1, s2) => comparison((T)v, s1, s2));
-
-        private void CreateSqlite3Collation(string name, object state, delegate_collation collation)
         {
-            var rc = raw.sqlite3_create_collation(_db, name, state, collation);
+            var rc = raw.sqlite3_create_collation(_db, name, state, (v, s1, s2) => comparison((T)v, s1, s2));
             SqliteException.ThrowExceptionForRC(rc, _db);
         }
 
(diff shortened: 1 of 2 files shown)


bricelam:
Per the [contributing guidelines](https://github.com/aspnet/Home/blob/dev/CONTRIBUTING.md):

> Before submitting a feature or substantial code contribution please discuss it with the team and ensure it follows the product roadmap.

Let's work together on a design before looking at any code. Could you add your proposal to the issue?

bricelam:
Re-opening. This is pretty close to the design we landed on. Would you mind updating it?

AlexanderTaeschner:
Updated to new API spec in #19.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -265,6 +265,30 @@ protected override DbCommand CreateDbCommand()
             => CreateCommand();
 
         /// <summary>
+        /// Create custom collation.
+        /// </summary>
+        /// <param name="name">Name of the collation.</param>
+        /// <param name="comparison">Method that compares two strings.</param>
+        public virtual void CreateCollation(string name, Comparison<string> comparison)
+            => CreateSqlite3Collation(name, null, (_, s1, s2) => comparison(s1, s2));
```
bricelam: Personally, I'd just call ` CreateCollation<object>(name, null, `... and inline `CreateSqlite3Collation` into the other overload.

AlexanderTaeschner: Changed.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -265,6 +265,27 @@ protected override DbCommand CreateDbCommand()
...
+        /// <summary>
+        /// Create custom collation.
+        /// </summary>
+        /// <typeparam name="T">The type of the state object.</typeparam>
+        /// <param name="name">Name of the collation.</param>
+        /// <param name="state">State object passed to each invokation of the collation.</param>
+        /// <param name="comparison">Method that compares two strings, using additional state.</param>
+        public virtual void CreateCollation<T>(string name, T state, Func<T, string, string, int> comparison)
+        {
+            var rc = raw.sqlite3_create_collation(_db, name, state, (v, s1, s2) => comparison((T)v, s1, s2));
```
bricelam: Oh, we'll need to throw if the connection isn't opened.

AlexanderTaeschner: Added check and test case.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -265,6 +265,32 @@ protected override DbCommand CreateDbCommand()
             => CreateCommand();
 
         /// <summary>
+        /// Create custom collation.
+        /// </summary>
+        /// <param name="name">Name of the collation.</param>
+        /// <param name="comparison">Method that compares two strings.</param>
+        public virtual void CreateCollation(string name, Comparison<string> comparison)
```
bricelam: Passing a `null` comparison will currently result in a `NullReferenceException`. We should either:
1. Throw `ArgumentNullException`
2. Support it by flowing `null` into `sqlite3_create_collation` (which removes the collation)

bricelam: We should probably  validate `name` too if the native API gives its "misuse" error for `null` or empty values.

AlexanderTaeschner: Opted for option 2 and added test.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
```diff
@@ -311,6 +312,67 @@ public void CreateCommand_returns_command()
...
+            {
+                connection.Open();
+                connection.CreateCollation("MY_NOCASE", (s1, s2) => string.Compare(s1, s2, StringComparison.OrdinalIgnoreCase));
+                connection.CreateCollation("MY_NOCASE", null);
+
+                var ex = Assert.Throws<SqliteException>(
+                    () => connection.ExecuteScalar<long>("SELECT 'Νικοσ' = 'ΝΙΚΟΣ' COLLATE MY_NOCASE;"));
+
+                Assert.Equal(raw.SQLITE_ERROR, ex.SqliteErrorCode);
+                Assert.Contains("no such collation sequence", ex.Message);
```
bricelam: Don't assert the message. (In case SQLite ever decides to localize.)

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -265,6 +265,33 @@ protected override DbCommand CreateDbCommand()
...
+        /// <param name="comparison">Method that compares two strings, using additional state.</param>
+        public virtual void CreateCollation<T>(string name, T state, Func<T, string, string, int> comparison)
+        {
+            if (State != ConnectionState.Open)
+            {
+                throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(CreateCollation)));
+            }
+
+            delegate_collation collation = comparison != null ? (v, s1, s2) => comparison((T)v, s1, s2) : (delegate_collation)null;
+            var rc = raw.sqlite3_create_collation(_db, name, state, collation);
```
bricelam: Do we need to validate `name`?

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
bricelam: Actually, I think there's a PRAGMA you can use to see the registered collations. It might be better to use that.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: Added validation.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
AlexanderTaeschner: Removed assertion. Adding a complete ExecuteReader loop for analyzing the PRAGMA seems to be overly complex. Since sqlite version 3.16.0 there are table-valued functions for the PRAGMAs usable in SELECTs, but I don't think that we can rely on the use of such a new version.

divega: OT: Are table-valued functions supported in general?

bricelam: @divega Nope. Only scalar arguments and one scalar result.

PR #343, opened 2017-03-24
Cleanup

bricelam:
The only functional change here is removing all GCed commands from the connection's weak reference list when one is explicitly removed.

cc @AlexanderTaeschner as FYI since this touches some of his recent PRs.


src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -212,39 +211,26 @@ protected override DbParameter CreateDbParameter()
...
         {
-            if (Connection == null
-                || Connection.State != ConnectionState.Open)
+            if (_connection?.State != ConnectionState.Open)
             {
                 throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(Prepare)));
             }
 
-            if (string.IsNullOrEmpty(CommandText))
+            if (string.IsNullOrEmpty(_commandText))
```
divega: Nit: Do we care to throw for white space?

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -212,39 +211,26 @@ protected override DbParameter CreateDbParameter()
...
-                            continue;
-                        }
-
-                        break;
-                    }
+                return;
+            }
 
-                    _preparedStatements.Add(stmt);
+            using (var enumerator = PrepareAndEnumerateStatements().GetEnumerator())
```
divega: Would `foreach` work? Is this better?

bricelam: It would work. This just avoid having a local variable for the current value. (It's the same thing we do for `.Load()`.)

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: It appears to be handled gracefully.

PR #346, opened 2017-03-24
Remove net451 as a cross-compile target

pranavkm:


src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
```diff
@@ -17,7 +17,7 @@ Microsoft.Data.Sqlite.SqliteException
 Microsoft.Data.Sqlite.SqliteFactory
 Microsoft.Data.Sqlite.SqliteParameter
 Microsoft.Data.Sqlite.SqliteTransaction</Description>
-    <TargetFrameworks>net451;netstandard1.2</TargetFrameworks>
+    <TargetFrameworks>net46;netstandard1.2</TargetFrameworks>
```
bricelam: So you'll get netstandard on net45x? That's no good...

bricelam: Better stick with the min net version supported by netstandard 1.2

davidfowl: why?

bricelam: Why cripple net45x? Targeting netstandard1.2 means this library supports it one way or another...

bricelam: By cripple, I mean: without crosstargeting calling DbDataReader.Close() is a no-op on net45x

bricelam: dotnet/corefx#2249

pranavkm: cc @Eilon

Eilon: The lowest planned .NET Framework target is currently .NET 4.6.1. I think if we have strong evidence to suggest it's worth targeting anything lower than that, let's bring it up with @DamianEdwards .

davidfowl: > https://github.com/dotnet/corefx/issues/2249

Wow...

bricelam:
Waiting on the outcome of dotnet/corefx#2249

PR #347, opened 2017-03-29
Converted test projects to run on netcoreapp2.0

kichalla:


test/Microsoft.Data.Sqlite.Tests/SqliteDataReaderTest.cs
```diff
@@ -693,8 +693,10 @@ public void IsClosed_returns_true_when_closed()
                 var reader = connection.ExecuteReader("SELECT 1;");
 #if NET46
                 reader.Close();
```
bricelam: In theory, this *should* work on `netcoreapp2.0` (but that will require cross-targeting `netstandard2.0` in the product)

PR #349, opened 2017-04-03
[WIP] Update Prepare implementation

### LINKED ISSUE #235: Implement SqliteCommand.Prepare

### LINKED ISSUE #345: Revert "Revert "Implement SqliteCommand.Prepare.""
Revert the PR #344 after investigating the downstream failures it caused. See https://github.com/aspnet/EntityFramework/issues/7982

bricelam:
Changes:
* Disallow multiple open readers for a single command
* Reset statements when finished (instead of before) reading

Resolves #235

Note, only https://github.com/aspnet/Microsoft.Data.Sqlite/pull/349/commits/68ca59bf10b00aa6e1e4261ad8d15134891edd71 needs review.

cc @AlexanderTaeschner


src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -180,14 +174,23 @@ protected override DbParameterCollection DbParameterCollection
...
+        /// <summary>
         /// Releases any resources used by the connection and closes it.
         /// </summary>
         /// <param name="disposing">
         /// true to release managed and unmanaged resources; false to release only unmanaged resources.
         /// </param>
         protected override void Dispose(bool disposing)
         {
-            DisposePreparedStatements();
+            if (disposing)
```
AlexanderTaeschner: I removed this condition before in order to enable the base finalizer to dispose the prepared statements.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -510,21 +518,19 @@ private IEnumerable<sqlite3_stmt> PrepareAndEnumerateStatements()
...
+        private void DisposePreparedStatements()
         {
-            foreach (var stmt in _preparedStatements)
+            if (DataReader != null)
             {
-                raw.sqlite3_reset(stmt);
-
-                yield return stmt;
+                DataReader.DisposeStatements = true;
+                DataReader = null;
```
AlexanderTaeschner: You are now depending on the Dispose method of the reader to be called to dispose the prepared statements. Before the base finalizer of the command did this everytime by calling Dispose(false).

AlexanderTaeschner:
You now readded a second list of the prepared statements to the reader, while we already got rid of one in the connection to have only one object responsible for disposing the statements. 
What about adding a wrapper around the sqlite3_stmt class which disposes the wrapped sqlite3_stmt instance when no other object (be it the reader or the command) needs it anymore?

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: This is an exceptional case that only happens when you do one of the following with an open SqliteDataReader
* Change the command's Connection
* Change the command's CommandText
* Dispose the command

All SqliteDataReader's should be explicitly disposed by the user per the ADO.NET contract.

If there is no active reader, disposing the command will dispose the statements.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: Before switching to SQLitePCL.raw, this case was handled by using `SafeHandle` objects for all the the native handles. All our code currently assumes this behavior. I'll follow up with a new issue to investigate how we need to react now that we're using SQLitePCL.raw

bricelam:
> You now readded a second list of the prepared statements to the reader

This list will almost always be empty. It also isn't really adding a new list but merely keeping a hold of all the dequeued statements until the reader is closed (instead of freeing them immediately after we're done with them).

This is, unfortunately, necessary to handle the exceptional case of a orphaned readers.

If it's really important, it the queue and the list could probably be collapsed into a single structure.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: Filed #351

AlexanderTaeschner:
I have nothing against the list itself, my only concern was adding an additional object which is responsibel for the cleanup of statements (although I see why it is necessary) shared by two objects. Hopefully the new issue #351 will help to clean this up later on.

AlexanderTaeschner:
Consider the following test case:
```
        [Fact]
        public void Command_can_be_disposed_when_open_reader()
        {
            {
                connection.Open();

                command.CommandText = "SELECT 1;";

                using (var reader = command.ExecuteReader())
                {
                    reader.Read();

                    command.Dispose();

                    Assert.Equal(1L, reader.GetValue(0));
                }
            }
        }
```
If you debug into the command.Dispose() call, you see that there is a prepared statement in `_preparedStatements` of the command, but the collection `_statementsToDispose` is empty. This statement is not disposed anymore.

bricelam:
The statement will still get disposed with the reader. _statementsToDispose, _stmt, and _stmtQueue all get disposed with the reader. Just using one list for everything might make this clearer.

bricelam:
A simpler approach I considered was to block changing CommandText or Connection and closing any open readers when the command is disposed. Will discuss with the team...

bricelam:
I will discuss throwing in `CommandText` and `Connection` at the design meeting tomorrow. This would remove the need to dispose statements in the data reader.

bricelam:
We decided to go with the simpler approach of throwing in the property setters and keeping the responsibility of freeing the statements just in SqliteCommand.

bricelam:
I'm still seeing failures on EntityFramework... ☹️

AlexanderTaeschner:
In my experiments with the EntityFramework tests and the prepare code I saw failures of the kind, that the database could not be deleted, because it was not closed. These would indicate, that some statements are not finalized before the call to the dispose of the Connection. From the documentation fo sqlite:

> If the database connection is associated with unfinalized prepared statements or unfinished sqlite3_backup objects then sqlite3_close() will leave the database connection open and return SQLITE_BUSY.

In the sqlite3.Dispose() routine from SQLitePCL.raw, the result of sqlite3_close is not checked. If you want to see, whether this is the reason of the failures, you can call
```csharp
            var rc = raw.sqlite3_close(_db);
            SqliteException.ThrowExceptionForRC(rc, _db);
```
instead of  `_db.Dispose()` in SqliteConnection.Dispose.

bricelam:
Thanks, that will certainly help with with debugging. I was sad to learn that SQLitePCL.raw didn't use `sqlite3_close_v2`. I still intend to track down where the leak is happening, but I've been stretched a bit thin this week...

AlexanderTaeschner:
If you want to dispose the sqlite3 objects of SQLitePCL.raw via sqlite3_close_v2(_db) instead of _db.Dispose() you can do it (it calls db.set_already_disposed() internally to mark the instance as disposed). But the v2 code does not help when statements are not disposed at all, only when they are disposed after the sqlite3_close_v2 call:
> If sqlite3_close_v2() is called with unfinalized prepared statements and/or unfinished sqlite3_backups, then the database connection becomes an unusable "zombie" which will automatically be deallocated when the last prepared statement is finalized or the last sqlite3_backup is finished.

bricelam:
Superseded by #367

Issue #14, opened 2014-04-02
User-defined functions


bricelam:
SQLite allows you to [create or redefine SQL functions](http://sqlite.org/c3ref/create_function.html). We want to enable this by using calls like the following.

```C#
connection.CreateFunction(
    "random",
    new Random(),
    r => r.Next());

connection.CreateFunction(
    "length",
    (string s) => s.Length);

connection.CreateFunction<long, long, long>(
    "max",
    Math.Max);

connection.CreateAggregate(
    "count",
    0,
    (long count, object[] args) => count++);

connection.CreateAggregate(
    "sum",
    (long sum, long i) => sum + i);
```

If you specify a parameter of type `object`, it will only ever be one of SQLite's primitive types: `long`, `double`, `string`, or `byte[]`.

If you specify a parameter of a type other than `object`, we will attempt to coerce the value into that type before passing it to you using the same semantics as `SqliteDataReader.GetFiledValue<T>()`.

The result will be coerced based on its type using the same semantics as `SqliteParameter.Value`.

If you specify a single parameter of type `object[]`, the function can take a variable number of arguments.

The definitions for the API would look like this. Methods with a type argument T1..N will be expanded into N overloads. We should generate the overloads using a `*.tt` file.

```C#
public class SqliteConnection
{
        public virtual void CreateFunction<TResult>(
            string name,
            Func<TResult> function);

        public virtual void CreateFunction<T1..16, TResult>(
            string name,
            Func<T1..16, TResult> function);

        public virtual void CreateFunction<TState, TResult>(
            string name,
            TState state,
            Func<TState, TResult> function);

        public virtual void CreateFunction<TState, T1..15, TResult>(
            string name,
            TState state,
            Func<TState, T1..15, TResult> function);

        public virtual TContext CreateAggregate<TContext>(
            string name,
            Func<TContext, TContext> func);

        public virtual TContext CreateAggregate<TContext, T1..15>(
            string name,
            Func<TContext, T1..15, TContext> func);

        public virtual TContext CreateAggregate<TContext>(
            string name,
            TContext seed,
            Func<TContext, TContext> func);

        public virtual TContext CreateAggregate<TContext, T1..15>(
            string name,
            TContext seed,
            Func<TContext, T1..15, TContext> func);

        public virtual TResult CreateAggregate<TContext, TResult>(
            string name,
            TContext seed,
            Func<TContext, TContext> func,
            Func<TContext, TResult> resultSelector);

[...]

Issue #354, opened 2017-04-06
Connection configuration


bricelam:
This was previously discussed in #61.

There are quite a few things in SQLite that need to be configured every time a connection is opened. (E.g. certain PRAGMAs, user-defined functions, collations, etc.)

The idea would be that you pass some configuration into your `SqliteConnection` instances, and every time the connection is opened, it would be configured appropriately.

AlexanderTaeschner:
For the PRAGMAs we could take the same route as System.Data.SQLite does it: use certain keywords in the ConnectionString. Looking at the ones from System.Data.SQLite one could go with:

- Foreign Keys: True, False
- Journal Mode: Delete, Truncate, Persist, Memory, Wal, Off
- Recursive Triggers: True, False
- Synchronous: Off, Normal, Full, Extra

Additional PRAGMA settings of interest could be:

- Auto Vacuum: None, Full, Incremental
- Automatic Index: True, False

For collations or user defined functions we would need an additional configuration object, but the information storage in this object will be tricky. One could add the same functions as defined in #14 and #19 inside this `SqliteConnectionConfiguration` class and use them to store the given information in lists which would be parsed later in the Open call of the collation.

bricelam:
I'm not a big fan of the things System.Data.Sqlite added to the connection string. In my mind, things in the connection string should be things you can't change after the connection is opened.

You ideas about `SqliteConnectionConfiguration` match mine.

```C#
var config = new SqliteConfiguration();
config.AddFunction("func" myDelegate);

var connection = new SqliteConnection(connectionString, config);

connection.Open(); // Automatically calls connection.AddFunction("func" myDelegate);
```

AlexanderTaeschner:
Okay, then the keywords mentioned above would be properties of the `SqliteConnectionConfiguration`.
Proposal would than be:
```
public class SqliteConnectionConfiguration
{
    public bool ForeignKeys {get; set;} = false; // default?
    public bool RecursiveTriggers {get; set;} = false; // default?
    public JournalMode JournalMode {get; set;} = JournalMode.Wal;// enum name? default?
    public SynchronousMode Synchronous {get; set;} = SynchronousMode.Full; // default?
    public AutoVacuumMode AutoVacuum {get; set;} = AutoVacuumMode.None; // default?
    public AutomaticIndex {get; set;} = true; // default?
    
    public void AddCollation(
        string name,
        Comparison<string> comparison);

    public void AddCollation<T>(
[...]

PR #355, opened 2017-04-10
Add creation of user defined functions

AlexanderTaeschner:
Add functions to create user defined functions.

Addresses #14


AlexanderTaeschner:
Ready to be reviewed (T4 was not so difficult with the excellent example given in #14 - thanks!!!).

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -265,6 +267,352 @@ protected override DbCommand CreateDbCommand()
...
+        {
+            delegate_function_aggregate_step stepFunction = null;
+            delegate_function_aggregate_final finalFunction = null;
+            if (function != null)
+            {
+                stepFunction = (ctx, _, args) =>
+                {
+                    if (ctx.state == null)
+                    {
+                        ctx.state = seed;
```
bricelam: If the result can be `null`, this may cause problems. One way around it would be to wrap the value in an type so we know if we've initialized it or not.
```C#
if (ctx.state == null)
    ctx.state = new MyHolder<TContext> { Value = seed };

ctx.state.Value = function(ctx.state.Value);
```

AlexanderTaeschner: Changed.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -297,6 +299,352 @@ public virtual void CreateCollation<T>(string name, T state, Func<T, string, str
...
+            if (type == typeof(DBNull))
+            {
+                return (T)(object)DBNull.Value;
+            }
+            if (type == typeof(object))
+            {
+                return (T)GetArgumentObject(value);
+            }
+
+            return SqliteDataReader.GetValue<T>(
```
bricelam: I don't like adding this responsibility directly to `SqliteDataReader`. I wonder if we can put all the type coercing into a base class (possibly `SqliteDataRecord` inheriting `DbDataRecord`), and create a separate derived class that provides APIs of a list of `sqlite3_value` objects.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -297,6 +299,352 @@ public virtual void CreateCollation<T>(string name, T state, Func<T, string, str
...
+            }
+
+            return array;
+        }
+
+        private void SetFunctionResult(object result, sqlite3_context ctx)
+        {
+            var type = result.GetType().UnwrapNullableType().UnwrapEnumType();
+
+            SqliteParameter.CoerceValue(
```
bricelam: Similarly, I don't like overloading `SqliteParameter`. We should also look into an abstraction over the concept of a `sqlite3_result` that shares a common base type with `SqliteParameter`.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: I tried the use of a class inheriting from DbDataRecord, but a large number of functions which have to be implemented for the IDataRecord interface, depend on the state of the DataReader, e.g.
```[csharp]
        public override bool IsDBNull(int ordinal)
            => _closed
                ? throw new InvalidOperationException(Resources.DataReaderClosed(nameof(IsDBNull)))
                : !_stepped || _done
                    ? throw new InvalidOperationException(Resources.NoData)
                    : GetSqliteType(ordinal) == raw.SQLITE_NULL;
```
I therefore created a new static class with all the value conversion.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: See above.

AlexanderTaeschner: With the newest version this is refactored into new classes which do the value storage. Since SqliteParameter uses this also for rebinding the value to a new statement/index the base class is a bit more complex than needed for the result storage.

bricelam:
@AlexanderTaeschner any chance you could rebase and squash this on top of `dev`?

AlexanderTaeschner:
Okay, rebased and squashed.

bricelam:
Sorry I'm taking forever on this one. I'm still not completely happy with the factoring. I'm going to play around with the code a bit and see if I can refine the parts I don't like a bit more. Forgive me; I'm an especially picky maintainer on this, [my weekend project](http://www.bricelam.net/2015/04/29/sqlite-on-corefx.html).

bricelam:
Just tested it, and throwing an exception "works", but it feels very incorrect since `sqlite3_step()` never returns. I'll update it to catch any exceptions and call `sqlite3_result_error()`.

AlexanderTaeschner:
Looking at issue #354 the question arrises if we should discard this PR and replace all Create... functions with Add... functions that work independent of the state of the connection (so that they can be issued even when the connection is not yet open).

AlexanderTaeschner:
As addition to Add.. functions maybe also add Remove... ones to delete functions.

bricelam:
#354 is going to require A LOT of design thought.

For now, I'd like to keep the programming model as close to the native APIs as possible. E.g. The connection needs to be open for most operations and they don't persist after the connection is closed.

ericsink:
Oops.  Sorry I missed a mention above.  I keep thinking I should pay more attention to this thread, because it looks interesting.  Anyway, it looks like you moved past that particular question you had for me.

bricelam:
Maybe in 3.0, we implement #354 not by adding new APIs, but by making the existing APIs persist. Whatever we do, I don't expect to tackle #354 any time soon.

bricelam:
@ericsink No worries. I was just checking to see if you've thought about it before. After testing I decided the safest thing to do was catch any exceptions and call `sqlite3_result_error()`.

bricelam:
There we go. There's noting a few more abstractions and levels of indirection can't solve. 😉

bricelam:
I need to add a few more tests around `null` (and `DBNull`) values, but the bulk of it is ready for review.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -377,5 +377,164 @@ public virtual void EnableExtensions(bool enable = true)
...
+            {
+                throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(CreateFunction)));
+            }
+
+            delegate_function_scalar func = null;
+            if (function != null)
+            {
+                func = (ctx, user_data, args) =>
+                {
+                    // TODO: Avoid allocation when niladic
```
bricelam: The TODOs left in this PR are theoretical perf improvements. I didn't want to make them without measuring the improvement since they would complicate the code base.

src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
```diff
@@ -208,137 +204,11 @@ internal bool Bind(sqlite3_stmt stmt)
...
-                else
-                {
-                    throw new InvalidOperationException(Resources.UnknownDataType(type));
-                }
-
-                _bindActionValid = true;
-            }
-
-            _bindAction(stmt, index);
+            new SqliteParameterBinder(stmt, index, _value).Bind();
```
AlexanderTaeschner: Why don't you store the binding action anymore? This will slow down the reuse of prepared statements,

bricelam: Over-optimization. We haven't done any performance profiling on the library, so we don't actually know what difference it makes. It would be nice to do an actual round of performance tuning this release.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
```diff
@@ -387,6 +388,397 @@ public void CreateCollation_works_with_state()
...
+        public void CreateFunction_works_when_parameter_null_but_type_not_nullable()
+        {
+            using (var connection = new SqliteConnection("Data Source=:memory:"))
+            {
+                connection.Open();
+                connection.CreateFunction("test", (long x) => x);
+
+                var result = connection.ExecuteScalar<long>("SELECT test(NULL);");
+
+                Assert.Equal(0L, result);
```
bricelam: This behavior is odd (when specifying NULL for non-nullable parameters), but it matches how SQLite will coerce NULL when you ask for a value of a specific type...

bricelam: The other option was to throw. I'll discuss with the team.

test/Microsoft.Data.Sqlite.Tests/SqliteDataReaderTest.cs
```diff
@@ -293,6 +305,23 @@ public void GetFieldValue_of_DBNull_works()
...
+            using (var connection = new SqliteConnection("Data Source=:memory:"))
+            {
+                connection.Open();
+
+                using (var reader = connection.ExecuteReader("SELECT 1;"))
+                {
+                    var hasData = reader.Read();
+
+                    Assert.True(hasData);
+                    Assert.Throws<InvalidCastException>(() => reader.GetFieldValue<DBNull>(0));
```
bricelam: In some extremely obscure case, one might be consider a breaking change.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateAggregate.cs
```diff
@@ -0,0 +1,897 @@
...
+        /// <typeparam name="T10">The type of the tenth parameter of the function.</typeparam>
+        /// <typeparam name="T11">The type of the eleventh parameter of the function.</typeparam>
+        /// <typeparam name="T12">The type of the twelfth parameter of the function.</typeparam>
+        /// <typeparam name="T13">The type of the thirteenth parameter of the function.</typeparam>
+        /// <typeparam name="T14">The type of the fourteenth parameter of the function.</typeparam>
+        /// <typeparam name="T15">The type of the fifteenth parameter of the function.</typeparam>
+        /// <typeparam name="TAccumulate">The type of the accumulator value.</typeparam>
+        /// <param name="name">The name of the SQL function.</param>
+        /// <param name="func">An accumulator function to be invoked on each element.  Pass null to delete a function.</param>
+        public virtual void CreateAggregate<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, TAccumulate>(string name, Func<TAccumulate, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, TAccumulate> func)
```
natemcmaster: Good grief that's a lot of type params. I hope this is never used.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateAggregate.tt
```diff
@@ -0,0 +1,134 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Data.Sqlite
+{
+    partial class SqliteConnection
+    {
+<#
+    var ordinal = new[] { "first", "second", "third", "fourth", "fifth", "sixth", "seventh", "eighth", "ninth", "tenth", "eleventh", "twelfth", "thirteenth", "fourteenth", "fifteenth" };
```
natemcmaster: Do we really need 15? `Action<T>` and `Func<T>` top out at 8.

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -153,6 +150,7 @@ public override bool NextResult()
             raw.sqlite3_reset(_stmt);
 
             (_stmt, _hasRows) = _stmtQueue.Dequeue();
+            _record = new SqliteDataRecord(_stmt);
```
natemcmaster: Have you profiled the perf impact of this? My concern is that this is the read hot path, and this adds an allocation per row, a potentially huge increase in memory traffic. Unless I missed it, SqliteDataRecord is only used from SqliteDataReader, which means it could stay inside this type and doesn't need to be factored out into a separate type.

bricelam: We've never profiled this product. We should.

This is only one allocation per result set (typically one per reader), not row. It shouldn't be significant.

No multiple inheritance; composition was the only option.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateAggregate.tt
bricelam: lol, I just used the max `Func` would allow on .NET Standard 2.0.

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
natemcmaster: > This is only one allocation per result set

Oh right. NextResult() means next result of a statement, not next row. Ok, my concern is gone.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateAggregate.tt
natemcmaster: Well, I guess it doesn't hurt.

test/Microsoft.Data.Sqlite.Tests/SqliteDataReaderTest.cs
natemcmaster: What's the case?

bricelam: `reader.GetFieldValue<DBNull>(0);` when the field is not `NULL`. Although I'd like to meet the person calling this with `DBNull`.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
bricelam: We decided to throw. I'll make the change.

Issue #273, opened 2016-06-18
Support saving GUID values to STRING columns


RonFrick:
Need a way to specify saving guid as text column. I would think adding a way to set BinaryGUID to false in the connection string would solve the issue.

natemcmaster:
@RonFrick this is already possible by calling `ToString()` when saving the guid. Did you have something else in mind?

Example:
```c#
                var guid = Guid.NewGuid();
                var save = conn.CreateCommand();
                save.CommandText = "INSERT INTO tbl VALUES (@p1)";
                save.Parameters.AddWithValue("@p1", guid.ToString());
                save.ExecuteNonQuery();
```

Full example: https://gist.github.com/natemcmaster/db14848ed53ae520e8a2b8a6c4410d52

RonFrick:
Hey thanks Nate! I am using EF core. Is there anyway to do the same thing using DbContext?

If not then I will certainly look a re writing my inserts manually and not use EF.

Thanks again for the help.

RonFrick:
Would there be something I can do in this class to always force it to save guids as text. 
Microsoft.Data.Sqlite.SqliteCommand

I already have my own version of this dll to solve other problems so I am fine for a one off fix on this issue as well. Any help is appreciated. Thanks Nate.

natemcmaster:
> Is there anyway to do the same thing using DbContext?

You would need to change your properties CLR type to System.String instead of System.Guid. EF wraps this layer with its own type mapping, but doesn't have any influence on how the CLR types are persisted by Microsoft.Data.Sqlite. 

Curious, why do you need to save as TEXT instead of the default ( which BLOB)?

> Would there be something I can do in this class to always force it to save guids as text. 

There are several different ways it could be done if you want the change the default behavior in your fork. One place to look is binding in SqliteParameter.cs.

RonFrick:
Ok...I will take a look and evaluate. Thank you!

The reason for needing to specify the behavior is because I am using a flavor of sync framework for sqlite here..
[](https://github.com/Mimetis/SyncWinRT)

The sync is saving GUIDs as text. EF is saving it as binary, but EF is also saving it in a format that cannot be queried again. After I save the row if you try to query that same row EF throws this... Guid should contain 32 digits with 4 dashes (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx). I can attach an image of what it looks like with SQLite browser. 

[...]

PR #363, opened 2017-04-28
Add BinaryGUID connection string to select binary/string GUID storage

AlexanderTaeschner:
Addresses #273.

To be able to access the new BinaryGUID property in ConnectionStringBuilder from the parameter binding I added a wrapper around the sqlite3_stmt object and add the associated connection to it.

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
@@ -219,18 +219,16 @@ namespace Microsoft.Data.Sqlite
             var changes = 0;
             var stmts = new Queue<(sqlite3_stmt, bool)>();
             var tail = CommandText;
+            int rc;
 
             do
             {
-                var rc = raw.sqlite3_prepare_v2(
-                        Connection.Handle,
-                        tail,
-                        out var stmt,
-                        out tail);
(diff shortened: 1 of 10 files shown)


bricelam:
This will greatly complicate any framework that generates SQL literals (including EFCore). I want to re-discuss with the team whether we should ever implement this feature.

AlexanderTaeschner:
Please keep in mind, that this only affects the way an GUID parameter is bind. Since #191 is resolved, the library can read both binary and string GUID without being told how the data is stored (thanks to the dynamic type system of sqlite).

bricelam:
Understood. Our philosophy so far has been to be highly compatible on reads, but highly predictable on writes. E.g. we support many formats when reading `DateTime`, `DateTimeOffset`, `Guid`, and `TimeSpan` values, but we will always write using a canonical format.

Currently, we'd generate queries like the following.
```SQL
SELECT $someGuidParam = x'SOME_BLOB_VALUE';
-- or
SELECT * FROM MyTable
WHERE SomeGuidColumn = x'SOME_BLOB_VALUE';
```
After this, those would return zero/empty, and we'd (EF and any other library using this) would need to change the format of `Guid` literals based on the connection string.

AlexanderTaeschner:
Ah okay, now I understand the problem.

bricelam:
Closing. We decided against taking this feature.

Issue #313, opened 2016-12-29
High CPU usage during SqliteCommand.ExecuteReader() on locked table


avolinsky:
100% CPU (on single core) usage while waiting for table to unlock during read.

### Minimal repro steps
1. Lock a table
2. Run select command against locked table
3. Observe CPU spike

i.e.

```
using (var connectionA = new SqliteConnection("Data Source=testdb;Mode=Memory;Cache=Shared")) {
				connectionA.Open();
				var cmd1 = connectionA.CreateCommand();
				cmd1.CommandText = "CREATE TABLE Person ( Name nvarchar(4000) )";
				cmd1.ExecuteNonQuery();

				using (var transactionA = connectionA.BeginTransaction(IsolationLevel.Serializable)) {
					cmd1.CommandText = "insert into Person values ( 'joe' )";
					cmd1.Transaction = transactionA;
					cmd1.ExecuteNonQuery();

					using (var connectionB = new SqliteConnection("Data Source=testdb;Mode=Memory;Cache=Shared")) {
						connectionB.Open();
						var cmd2 = connectionB.CreateCommand();
						cmd2.CommandText = "select * from  Person";
						cmd2.ExecuteReader(); //high CPU until timeout
					}
				}
			}
```

### Expected result
Minimal CPU usage while waiting for table to unlock

### Actual result
Waiting takes all available CPU cycles

### Further technical details
Missing a sleep [in this loop](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/SqliteCommand.cs#L283) ?

bricelam:
For triage context, the reason we didn't put a sleep there is because there is no Thread class on .NET Standard 1.3.

bricelam:
Will update with the move to .NET Standard 2.0.

PR #370, opened 2017-05-16
Target .NET Standard 1.4 & 2.0

### LINKED ISSUE #350: Cross-target netstandard2.0
To work around dotnet/corefx#2249 on netstandard2.0 and netcoreapp2.0, we need to cross-target netstandard2.0 (like we do net451)

bricelam:
Ideally, we'd just target .NET Standard 2.0, but since it will take a few months for UWP to support it, we'll cross-target 1.4 in the mean time. (See https://github.com/aspnet/Microsoft.Data.Sqlite/pull/370/commits/64319deb873fea6df8d9903ce3fd4997a07a89f9 for the code we'll be able to remvoe.)

Blocked waiting on updates to KoreBuild

Resolves #350

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
+++ b/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
@@ -17,7 +17,7 @@ Microsoft.Data.Sqlite.SqliteException
 Microsoft.Data.Sqlite.SqliteFactory
 Microsoft.Data.Sqlite.SqliteParameter
 Microsoft.Data.Sqlite.SqliteTransaction</Description>
-    <TargetFrameworks>net451;netstandard1.2</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <CodeAnalysisRuleSet>Microsoft.Data.Sqlite.Core.ruleset</CodeAnalysisRuleSet>
     <IncludeSymbols>true</IncludeSymbols>
@@ -32,11 +32,7 @@ Microsoft.Data.Sqlite.SqliteTransaction</Description>
 
   <ItemGroup>
     <PackageReference Include="SQLitePCLRaw.core" Version="$(SQLitePCLRawVersion)" />
-    <PackageReference Include="StyleCop.Analyzers" Version="$(StyleCopAnalyzersVersion)">
-      <PrivateAssets>All</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="System.Data.Common" Version="$(CoreFxVersion)" />
-    <PackageReference Include="System.ValueTuple" Version="$(CoreFxVersion)" />
+    <PackageReference Include="StyleCop.Analyzers" Version="$(StyleCopAnalyzersVersion)" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
+++ b/src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
@@ -17,7 +17,7 @@ Microsoft.Data.Sqlite.SqliteException
 Microsoft.Data.Sqlite.SqliteFactory
 Microsoft.Data.Sqlite.SqliteParameter
 Microsoft.Data.Sqlite.SqliteTransaction</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netstandard1.4</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <CodeAnalysisRuleSet>Microsoft.Data.Sqlite.Core.ruleset</CodeAnalysisRuleSet>
     <IncludeSymbols>true</IncludeSymbols>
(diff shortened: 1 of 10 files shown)


davidfowl:
Why both?

NuGet.config
```diff
@@ -4,5 +4,7 @@
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <!-- TODO: Remove -->
+    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
```
davidfowl: Remove before commit right?

bricelam: Yes, just waiting for us to take the latest build...

bricelam:
> Why both?

UWP

alexvaluyskiy:
@bricelam @davidfowl Wait wait, are you planning to remove support for .NET 4.5.1 (and NetStandard 1.4 eventually) without any discussing with the community? Again? Only to have a posibility to remove a couple of lines of code? It is not serious. I think it is the same issue as MS had here: https://github.com/aspnet/Home/issues/2022

`Microsoft.Data.Sqlite` is only one ADO.NET Sqlite provider which supports Linux and much faster that the official `System.Data.SQLite.Core` package. It has more than 400_000 downloads on [nuget](https://www.nuget.org/packages/Microsoft.Data.Sqlite/2.0.0-preview1-final) and could be used outside of MS Stack (ASP.NET Core, Entity Framework).
The most of developers still want to have `net4x` support, why you want to remove it?

willdean:
@alexvaluyskiy There doesn't seem to be any discussion here about dropping .NET standard (which includes full netfx) - this is about changing to .NET standard 2.0.  It's not the same thing as the notorious 2022 thread.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
```diff
@@ -17,7 +17,7 @@ Microsoft.Data.Sqlite.SqliteException
 Microsoft.Data.Sqlite.SqliteFactory
 Microsoft.Data.Sqlite.SqliteParameter
 Microsoft.Data.Sqlite.SqliteTransaction</Description>
-    <TargetFrameworks>net451;netstandard1.2</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard1.4</TargetFrameworks>
```
alexvaluyskiy: Why have you removed `net451`?

alexvaluyskiy:
@willdean Look at the changes. They have removed support for `net451`

MartinJohns:
@alexvaluyskiy They're replacing it with `netstandard2.0`. .NET Standard 2.0 is supported by .NET 4.6.1, so .NET support is not removed.

schotime:
Yeh. Netstandard 2.0 is net461 min. This is not the same thing.

src/Microsoft.Data.Sqlite.Core/Microsoft.Data.Sqlite.Core.csproj
willdean: Because netstandard2.0 implies it.

MartinJohns: Because it's being replaced with .NET Standard 2.0, which is supported by .NET 4.6.1.

davidfowl:
@alexvaluyskiy across the board we don't plan to support older .NET Framework versions. .NET Framework 4.6.1 will be required in most if not all cases to use the newer versions of these libraries.

alexvaluyskiy:
@davidfowl I don't have any problems with net461. Just to announce it, that beforehand. Developers and the authors of libraries should know MS strategy. 
Which one of frameworks they should choose to migration in the future: 4.5.2 (which still supported by MS) or .NET 4.6.1. @davidfowl I don't have any problems with net461. Just to announce it, that beforehand. Developers and the authors of libraries should know MS strategy. 
Which one of frameworks they should choose to migration in the future: 4.5.2 (which still supported by MS) or .NET 4.6.1. It is no that clear at the moment

davidfowl:
@alexvaluyskiy We officially announced that we're going to *explicitly* target .NET Standard 2.0 on the blog post *and* in the announcements repository https://github.com/aspnet/Announcements/issues/243.

yaakov-h:
@alexvaluyskiy Almost all of the latest round of Microsoft libraries require .NET 4.6 or higher. Given that .NET Standard 1.4-2.0 all require 4.6.1 or higher (with tooling 2.0), and the announcement linked above, it's no surprise that you should be on or moving towards 4.6.1, 4.6.2, or 4.7.

alexvaluyskiy:
@davidfowl is it mean `net461` TFM or `netstandard2.0` which support net461?

davidfowl:
The announcement literally says this:

image

The blog post says "Our goal is to ship ASP.NET Core 2.0 on .NET Standard 2.0 so applications can run on .NET Core, Mono and .NET Framework". The announcement tries to go into more details but it really is all there...

alexvaluyskiy:
> Retarget all Microsoft.AspNetCore.* packages to netstandard2.0. 
> Cross-compiling to net461 when APIs are only available in net461 and netcoreapp2.0

That's mean that all assemblies will target `netstandard2.0` by default. What about Visual Studio 2015 Update 3 with nuget 3.4.4? I can't use any `netstandard2.0` packages from `net461` project at the moment. Do you plan to make the update of the tooling for VS 2015?

alexvaluyskiy:
Also, all projects from `Microsoft.Extensions.*` still target `netstandard1.x`, because **use the lowest version of netstandard if you can**. Why do you want to drop `netstandard1.4` here in the future (or even right now)?

bricelam:
> Why do you want to drop netstandard1.4 here

We view the library as a bit crippled on .NET Standard 1.x. A good example is #313. Because `Thread` didn't exist before .NET Standard 2.0, we had no way to yield in the retry loop leading to unnecessary CPU spikes.

gulbanana:
moving from .net 4.5 to 4.6 is a nonissue for almost everybody. the change is fine, being able to do this sort of thing was the entire goal of ns2.0

PR #386, opened 2017-07-13
Dispose all the things

bricelam:
(may want to view [without whitespace changes](https://github.com/aspnet/Microsoft.Data.Sqlite/pull/386/files?w=1))

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/Extensions/SQLitePCLExtensions.cs
+++ b/src/Microsoft.Data.Sqlite.Core/Extensions/SQLitePCLExtensions.cs
@@ -1,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace SQLitePCL
+{
+    internal static class SQLitePCLExtensions
+    {
+        public static void Dispose2(this sqlite3 db)
+            => raw.sqlite3_close_v2(db);
+    }
+}
--- a/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
@@ -220,7 +220,7 @@ namespace Microsoft.Data.Sqlite
             }
 
             Transaction?.Dispose();
-            _db.Dispose();
+            _db.Dispose2();
             _db = null;
             SetState(ConnectionState.Closed);
         }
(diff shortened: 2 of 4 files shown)


src/Microsoft.Data.Sqlite.Core/Extensions/SQLitePCLExtensions.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace SQLitePCL
+{
+    internal static class SQLitePCLExtensions
+    {
+        public static void Dispose2(this sqlite3 db)
```
ajcvickers: Dispose2? :trollface:

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -220,7 +220,7 @@ public override void Close()
             }
 
             Transaction?.Dispose();
-            _db.Dispose();
```
ajcvickers: So no call to the "real" Dispose at all?

ericsink:
Looks like Brice is (more or less) working around the lack of finalizers in my library.  

Ref: ericsink/SQLitePCL.raw#166

Brice has also (more or less) convinced me that I need to address this issue.

When SQLitePCL.raw goes to 2.0, I will probably fix this by inheriting from `SafeHandle`.  But I do think I want to go ahead and deal with this in 1.x with finalizers.

Obviously this is a simple task, but I've still wanted to review all the standard stuff on finalizers and dispose and SuppressFinalize, trying to make sure I get it right.

Just an FYI here...

bricelam:
Sweet, I'll be sure to remove any unnecessary code here when we update our SQLitePCL.raw version.

src/Microsoft.Data.Sqlite.Core/Extensions/SQLitePCLExtensions.cs
bricelam: Underneath, we switched from calling `sqlite3_close` to `sqlite3_close_v2`. *Dispose2* seemed appropriate 😉

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: We could, but it would be a no-op since `raw.sqlite3_close_v2` marks the `sqlite3` object as already disposed.

PR #388, opened 2017-07-20
Update SqliteCommand.Prepare implementation

### LINKED ISSUE #235: Implement SqliteCommand.Prepare

bricelam:
https://github.com/aspnet/Microsoft.Data.Sqlite/pull/388/commits/8b37cabb5214eab12cba49fd915d965c3bad9ca0 improves on the previous implementation we tried.

I ran a quick benchmark reusing a parameterized `INSERT` command for fun: 4x faster!

&nbsp; | Mean | Error | StdDev
--- | ---:| ---:| ---:
**Before** | 4.184 us | 0.0835 us | 0.2001 us
**After** | 1.048 us | 0.0208 us | 0.0415 us

Resolves #235 

/cc @AlexanderTaeschner


src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -282,38 +359,35 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
...
-                    //     It will result in false positives, but it's the best we can do without re-parsing SQL
-                    || raw.sqlite3_stmt_readonly(stmt) != 0)
-                {
-                    stmts.Enqueue((stmt, rc != raw.SQLITE_DONE));
                 }
-                else
+            }
+            catch
+            {
+                if (unprepared)
```
bricelam: Will update to `catch..when`.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -76,13 +79,48 @@ public override CommandType CommandType
...
+        {
+            get => _commandText;
+            set
+            {
+                if (DataReader != null)
+                {
+                    throw new InvalidOperationException(Resources.SetRequiresNoOpenReader(nameof(CommandText)));
+                }
+
+                if (!value.Equals(_commandText))
```
bricelam: possible null ref. Will update.

src/Microsoft.Data.Sqlite.Core/Properties/Resources.resx
```diff
@@ -198,4 +198,10 @@
   <data name="CannotStoreNaN" xml:space="preserve">
     <value>Cannot store 'NaN' values.</value>
   </data>
+  <data name="DataReaderOpen" xml:space="preserve">
+    <value>An open DataReader is already associated with this command. Close it before opening a new one.</value>
```
bricelam: Should be *data reader*.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -441,5 +515,51 @@ public override object ExecuteScalar()
...
+
+        private IEnumerable<sqlite3_stmt> PrepareAndEnumerateStatements()
+        {
+            var tail = _commandText;
+            do
+            {
+                var rc = raw.sqlite3_prepare_v2(
+                    _connection.Handle,
+                    tail,
+                    out sqlite3_stmt stmt,
```
bricelam: `var`

Issue #394, opened 2017-08-12
Broken reuse of SqliteCommand and SqliteParameter


zone117x:
An initial SqliteCommand and SqliteParameters cannot be used through multiple query executions. 
Only the first value of the sqlite parameter(s) are used. The workaround is to reset the SqliteCommand.CommandText which causes the prepared statements to be cleared.

I've forked and added a unit test method that demonstrates the bug with the hacky workaround.

https://github.com/PingmanTools/Microsoft.Data.Sqlite/blob/prepared-stmt-bug/test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs#L129

TLDR: The following doesn't work as expected:

```cs
using (var command = connection.CreateCommand())
{
    command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
    var valueParam = new SqliteParameter { ParameterName = "@value" };
    command.Parameters.Add(valueParam);

    for (var i = 2; i < 5; i++)
    {
        command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
        valueParam.Value = i;
        command.ExecuteNonQuery();
    }
}
```
All inserted rows will have `Value` set to `2` (the first set value of the parameter).

bricelam:
@zone117x What version are you using? The nightly builds?

zone117x:
Latest dev branch. As shown by the fork with the added unit test

bricelam:
My subconscious [was trying to tell me](https://github.com/aspnet/Microsoft.Data.Sqlite/pull/334#pullrequestreview-27627499) there might be bugs in this area. I just forgot to follow up. 😄

bricelam:
Weird. The following works:
```csharp
using (var command = connection.CreateCommand())
{
    command.CommandText = "SELECT @value;";
    var parameter = new SqliteParameter { ParameterName = "@value" };
    command.Parameters.Add(parameter);

    for (var i = 2L; i < 5L; i++)
    {
        parameter.Value = i;

        var result = command.ExecuteScalar();

        Assert.Equal(i, result);
    }
}
```

PR #396, opened 2017-08-16
Fix broken reuse of SqliteCommand

AlexanderTaeschner:
According to http://sqlite.org/c3ref/stmt.html the life cycle of a prepared statement looks like this:

1. Create the prepared statement object using sqlite3_prepare_v2(). 
2. Bind values to parameters using the sqlite3_bind_*() interfaces. 
3. Run the SQL by calling sqlite3_step() one or more times. 
4. Reset the prepared statement using sqlite3_reset() then go back to step 2. Do this zero or more times. 
5. Destroy the object using sqlite3_finalize(). 

Since we now store and reuse prepared statements we have to reset them before rebinding (4->2).
This reset got lost somewhere in the process of adding the Prepare() command.

Addresses #394

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
@@ -322,6 +322,11 @@ namespace Microsoft.Data.Sqlite
                     ? PrepareAndEnumerateStatements()
                     : _preparedStatements)
                 {
+                    if (!unprepared)
+                    {
+                        raw.sqlite3_reset(stmt);
+                    }
+
                     var boundParams = 0;
 
                     if (_parameters.IsValueCreated)
--- a/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
@@ -322,11 +322,6 @@ namespace Microsoft.Data.Sqlite
                     ? PrepareAndEnumerateStatements()
                     : _preparedStatements)
                 {
-                    if (!unprepared)
-                    {
-                        raw.sqlite3_reset(stmt);
-                    }
-
                     var boundParams = 0;
 
                     if (_parameters.IsValueCreated)
@@ -378,6 +373,7 @@ namespace Microsoft.Data.Sqlite
                     }
                     else
                     {
+                        raw.sqlite3_reset(stmt);
                         hasChanges = true;
                         changes += raw.sqlite3_changes(_connection.Handle);
                     }
--- a/test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
+++ b/test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
@@ -178,15 +178,12 @@ namespace Microsoft.Data.Sqlite
             using (var connection = new SqliteConnection("Data Source=:memory:"))
             {
                 connection.Open();
+                connection.ExecuteNonQuery("CREATE TABLE Data (ID integer PRIMARY KEY, Value integer);");
 
                 using (var command = connection.CreateCommand())
                 {
-                    command.CommandText = "CREATE TABLE Data (ID integer PRIMARY KEY, Value integer);";
-                    command.ExecuteNonQuery();
-
                     command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
-                    var valueParam = new SqliteParameter { ParameterName = "@value", Value = -1 };
-                    command.Parameters.Add(valueParam);
+                    var valueParam = command.Parameters.AddWithValue("@value", -1);
 
                     Assert.Equal(1, command.ExecuteNonQuery());
 
--- a/test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
+++ b/test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
@@ -171,6 +171,63 @@ namespace Microsoft.Data.Sqlite
         }
 
         [Fact]
+        public void Multiple_command_executes_works()
+            => Multiple_command_executes_works((SqliteCommand command) =>
+            {
+                command.CommandText = string.Empty;
+                command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
+            });
+
+        [Fact]
+        public void Multiple_command_executes_works_2()
+            => Multiple_command_executes_works((SqliteCommand command) =>
+            {
+                command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
+            });
+
+        [Fact]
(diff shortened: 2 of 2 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -322,6 +322,11 @@ public new virtual SqliteDataReader ExecuteReader(CommandBehavior behavior)
                     ? PrepareAndEnumerateStatements()
                     : _preparedStatements)
                 {
+                    if (!unprepared)
+                    {
+                        raw.sqlite3_reset(stmt);
```
bricelam: Let's do this in `SqliteDataReader` when `sqlite3_reset` is called.

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
```diff
@@ -171,6 +171,63 @@ public void Prepare_throws_when_command_text_contains_dependent_commands()
...
+
+                    for (var i = 0; i < INSERTS; i++)
+                    {
+                        action(command);
+
+                        valueParam.Value = i;
+                        Assert.Equal(1, command.ExecuteNonQuery());
+                    }
+
+                    command.CommandText = "SELECT Value FROM Data";
```
bricelam: Add `ORDER BY Value` since the order is technically undefined otherwise.

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
```diff
@@ -171,6 +171,63 @@ public void Prepare_throws_when_command_text_contains_dependent_commands()
...
+        [Fact]
+        public void Multiple_command_executes_works_2()
+            => Multiple_command_executes_works((SqliteCommand command) =>
+            {
+                command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
+            });
+
+        [Fact]
+        public void Multiple_command_executes_works_3()
+            => Multiple_command_executes_works((SqliteCommand command) => { });
```
bricelam: I think just this variation is needed. The other two aren't really testing anything related to the parameter binding.

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
```diff
@@ -171,6 +171,63 @@ public void Prepare_throws_when_command_text_contains_dependent_commands()
...
+
+                    command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
+                    var valueParam = new SqliteParameter { ParameterName = "@value" };
+                    command.Parameters.Add(valueParam);
+
+                    for (var i = 0; i < INSERTS; i++)
+                    {
+                        action(command);
+
+                        valueParam.Value = i;
```
bricelam: We may also want to add a test that using the same parameter value (i.e. not setting `SqliteParameter.Value`) re-uses the value for each execution.

bricelam:
P.S. Thanks for going above and beyond and sending a PR for this!

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
AlexanderTaeschner: In case of INSERTs (as in the test case) the DataReader has no access to the statement, since we only read back the value of changed rows (passed queue is null).

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
AlexanderTaeschner: I added two additional INSERTs in the test case.

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
AlexanderTaeschner: Added ORDER BY ID to also test the sequence of INSERTs.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
bricelam: Oooooooh, I see now. For some reason I assumed this call was `sqlite3_reset_bindings()`. I should actually look at the code when reviewing. 😉

In that case, and we do this when we decide not to pass it to the reader?

bricelam: (My goal here is to reset it as soon as we're done with it rather than just before we need to re-use it.)

AlexanderTaeschner: Okay I moved the reset behind the stepping (statements passed to the reader are already being reset).

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
```diff
@@ -171,6 +171,53 @@ public void Prepare_throws_when_command_text_contains_dependent_commands()
...
+                connection.Open();
+
+                using (var command = connection.CreateCommand())
+                {
+                    command.CommandText = "CREATE TABLE Data (ID integer PRIMARY KEY, Value integer);";
+                    command.ExecuteNonQuery();
+
+                    command.CommandText = "INSERT INTO Data (Value) VALUES (@value);";
+                    var valueParam = new SqliteParameter { ParameterName = "@value", Value = -1 };
+                    command.Parameters.Add(valueParam);
```
bricelam: Can simplify to...
```csharp
var valueParam = command.Parameters.AddWithValue("@value", -1);
```

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
```diff
@@ -171,6 +171,53 @@ public void Prepare_throws_when_command_text_contains_dependent_commands()
...
+            const int INSERTS = 3;
+
+            using (var connection = new SqliteConnection("Data Source=:memory:"))
+            {
+                connection.Open();
+
+                using (var command = connection.CreateCommand())
+                {
+                    command.CommandText = "CREATE TABLE Data (ID integer PRIMARY KEY, Value integer);";
+                    command.ExecuteNonQuery();
```
bricelam: Can simplify to...
```csharp
connection.ExecuteNonQuery("CREATE TABLE Data (ID integer PRIMARY KEY, Value integer);");
```

AlexanderTaeschner: Done.

test/Microsoft.Data.Sqlite.Tests/SqliteCommandTest.cs
AlexanderTaeschner: Done.

Issue #253, opened 2016-05-18
Implement IDbColumnSchemaGenerator


natemcmaster:
System.Data.Common 4.1.0 adds support for an interface that allows pulling metadata out of `DbDataReader`. https://github.com/dotnet/corefx/issues/5915

In short, this means implementing `System.Data.Common.IDbColumnSchemaGenerator` on `SqliteDataReader`

Cref https://github.com/dotnet/corefx/pull/8639.

saurabh500:
@natemcmaster It is not necessary for you to implement `System.Data.Common.IDbColumnSchemaGenerator` on `SqliteDataReader`  unless you want to expose the schema from the `SqliteDataReader`

divega:
Thanks @saurabh500. I think the point of this issue is that to consider doing it in the future.

bricelam:
See also #16

bricelam:
Note, in .NET Standard 2.0, it looks like `DbDataReader.GetColumnSchema()` delegates to `DbDataReader.GetSchemaTable()` if `IDbColumnSchemaGenerator` isn't implemented.

In other words, just implementing `DbConnection.GetSchema()` would be the best way to resolve this issue.

tiandian:
Please implementing DbDataReader.GetSchemaTable.
My below code doesn't work because sqlite's DbDataReader.GetSchemaTable doesn't implement.

```
DataTable dt = new DataTable();
using (DbDataReader ddr = this.ExecuteReader(strSql))
 { dt.Load(ddr); }
```
Below is the error:
```
System.NotSupportedException occurred
  HResult=0x80131515
  Message=Specified method is not supported.
  Source=<Cannot evaluate the exception source>
  StackTrace:
   at System.Data.Common.DbDataReader.GetSchemaTable()
   at System.Data.ProviderBase.SchemaMapping..ctor(DataAdapter adapter, DataSet dataset, DataTable datatable, DataReaderContainer dataReader, Boolean keyInfo, SchemaType schemaType, String sourceTableName, Boolean gettingData, DataColumn parentChapterColumn, Object parentChapterValue)
   at System.Data.Common.DataAdapter.FillMappingInternal(DataSet dataset, DataTable datatable, String srcTable, DataReaderContainer dataReader, Int32 schemaCount, DataColumn parentChapterColumn, Object parentChapterValue)
[...]

PR #397, opened 2017-08-17
Add initial implementation of GetSchemaTable

AlexanderTaeschner:
Add initial implementation of DbDataReader.GetSchemaTable.

This is based on the table in #253. The null entries are commented out at the moment. For the sequence of columns I used the same order as in SqlDataReader (https://referencesource.microsoft.com/#System.Data/System/Data/SqlClient/SqlDataReader.cs,451).

sqlite3_table_column_metadata is not supported by SQLitePCL.raw yet, so that IsKey, AllowDBNull and IsAutoIncrement are not implemented.

Addresses #253

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
@@ -533,7 +533,7 @@ namespace Microsoft.Data.Sqlite
                 schemaRow[IsExpression] = raw.sqlite3_column_origin_name(_stmt, i) == null;
                 //schemaRow[IsRowVersion] = null;
                 //schemaRow[IsHidden] = null;
-                schemaRow[IsLong] = GetFieldType(i) == typeof(long);
+                schemaRow[IsLong] = DBNull.Value;
                 //schemaRow[IsReadOnly] = null;
                 //schemaRow[ProviderSpecificDataType] = null;
                 //schemaRow[NonVersionedProviderType] = null;
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
@@ -448,26 +448,20 @@ namespace Microsoft.Data.Sqlite
             var schemaTable = new DataTable("SchemaTable");
 
             var ColumnName = new DataColumn(SchemaTableColumn.ColumnName, typeof(string));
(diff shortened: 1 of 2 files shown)


AlexanderTaeschner:
Looking at the source code of SQLitePCL.raw I see that sqlite3_table_column_metadata seems to be included. I will take a second look at it later this day (intellisense didn't show it).

bricelam:
Eric may not have included `sqlite3_table_column_metadata` in his API.

We can use this instead:
```sql
SELECT pk, notnull
FROM pragma_table_info($table)
WHERE name = $column
```

bricelam:
...and autoincrement:
```sql
SELECT COUNT(*)
FROM pragma_table_info($table)
WHERE name = $column
    AND pk = 1
    AND NOT EXISTS(
        SELECT *
        FROM pragma_index_list($table)
        WHERE origin = 'pk');
```

AlexanderTaeschner:
Seems that I was simply blind. I added now the call to sqlite3_table_column_metadata and fill IsKey, AllowDBNull, and IsAutoIncrement.

bricelam:
Oh man, I was so proud of my queries too...

bricelam:
Could you rebase this now that #355 is merged?

AlexanderTaeschner:
Rebased it without the possibility to test it locally (see #413).

AlexanderTaeschner:
Added proposed result (null) for IsLong.

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -436,5 +438,139 @@ public override object GetValue(int ordinal)
...
+                var columnName = raw.sqlite3_column_origin_name(_stmt, i);
+                schemaRow[BaseColumnName] = columnName;
+                //schemaRow[BaseSchemaName] = null;
+                var tableName = raw.sqlite3_column_table_name(_stmt, i);
+                schemaRow[BaseTableName] = tableName;
+                schemaRow[DataType] = GetFieldType(i);
+                schemaRow[DataTypeName] = GetDataTypeName(i);
+                //schemaRow[ProviderType] = null;
+                schemaRow[IsAliased] = raw.sqlite3_column_origin_name(_stmt, i) != GetName(i);
+                schemaRow[IsExpression] = raw.sqlite3_column_origin_name(_stmt, i) == null;
```
bricelam: Use `columnName` for these two calls to `sqlite3_column_origin_name()`.

AlexanderTaeschner:
Okay, changed as requested.

PR #399, opened 2017-08-18
Use Directory.Build.props/targets

natemcmaster:
Using more MSBuild magic :). This repo is small enough it doesn't make a big difference, but in bigger repos, it helps us reduce duplication between csproj files in src/ and test/ folders.

Also, Directory.Build.props is required if we every decided to change properties like BaseIntermediateOutputPath.


Microsoft.Data.Sqlite.sln
```diff
@@ -13,6 +13,25 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.Sqlite.Tests
...
+		CONTRIBUTING.md = CONTRIBUTING.md
+		Directory.Build.props = Directory.Build.props
+		Directory.Build.targets = Directory.Build.targets
+		LICENSE.txt = LICENSE.txt
+		NuGet.config = NuGet.config
+		NuGetPackageVerifier.json = NuGetPackageVerifier.json
+		README.md = README.md
+		version.xml = version.xml
+	EndProjectSection
+EndProject
```
bricelam: 🤢 do not want

natemcmaster: Just remove Solution items altogether? Or do you only want a subset of these files? I don't care much either way.

Directory.Build.targets
```diff
@@ -0,0 +1,6 @@
+<Project>
+    <!--
+        This file is intentionally included even thought it is blank.
```
smitpatel: Typo though => thought

Microsoft.Data.Sqlite.sln
smitpatel: If they are not useful for solution then should just remove all of them.

smitpatel:
Just side question - how does VS import things in project now?

Microsoft.Data.Sqlite.sln
natemcmaster: I'll let you tell me what you think is useful. IMO they all are useful files, but I rarely work on this project so I'll defer to what you guys want.

natemcmaster:
By default, MSBuild searches upwards to find the first Directory.Build.props/targets file, starting in the project directory. If it finds a file, it imports it automagically.

Microsoft.Data.Sqlite.sln
smitpatel: I am failing to understand but why VS needs to know about README.md? Does it show up in solution explorer i.e. VS light up something based on mention of the file. If putting file here has no impact on VS experience then they would be un-useful to be mentioned in solution file.

natemcmaster: "Solution Items" is a way of having generic files show up in the solution explorer. Their presence in the sln file doesn't affect the build. Makes it easier to double-click and open a file for editing if you use VS for everything.

image
.

smitpatel: Thanks for info.VS lighting up and making editing easier is good to have though its also clutter because those files are hardly changed or artifacts of different build systems we have (like appveyor/travis).
You can merge this PR with files included in sln. We'll discuss this in EF team and make a decision.

Issue #357, opened 2017-04-17
Implement Data Change Notifications Callback


allanrsmith:
### Data Change Notifications Callback
Unable to register a callback on the SqliteConnection object for updates to specified tables

### Functional impact
With separate App/DbContext instances I would like callbacks to be called in a second instance when one instance updates a table row.  This will allow for performant responsive applications.

### Minimal repro steps
Functionality is missing

### Expected result
Register a callback on the SqliteConnection for a specific table/set of tables, perhaps for types of table changes, to allow my code to react to table updates.

### Actual result
No functionality exists to support this

### Further technical details
https://sqlite.org/c3ref/update_hook.html

CC: @bricelam

bricelam:
Since we expose the native handle, one workaround could be:
```C#
sqlite3_update_hook(
    connection.Handle,
    (user_data, type, database, table, rowid) => { /* TODO: Handle it. */ },
    null);
```

AlexanderTaeschner:
What about the following addition to SqliteConnection:

```[csharp]
public event SqliteDataChangeEventHandler DataChange;

protected virtual void OnDataChange(SqliteDataChangeEventArgs e)
```

with

```[csharp]
public delegate void SqliteDataChangeEventHandler(object sender, SqliteDataChangeEventArgs e);

public class SqliteDataChangeEventArgs : EventArgs
{
    public SqliteDataChangeEventArgs(
        DataChangeAction dataChangeAction,
        string database,
        string table,
        long rowid)

    public DataChangeAction DataChangeAction { get; }
    public string Database { get; }
    public string Table { get; }
    public long Rowid { get; }
}

public enum DataChangeAction
{
    Unknown,
    Update,
    Delete,
    Insert
}
```

See https://github.com/AlexanderTaeschner/Microsoft.Data.Sqlite/commit/a3b0938cd42c5207723c01c0be1c54aa661afdb5 for a test implementation.

bricelam:
Does System.Data.SQLite expose any API for it?

AlexanderTaeschner:
I didn't look at System.Data.SQLite before I made my suggestion. System.Data.SQLite.SQLiteConnection has the following event:
```[csharp]
public event SQLiteUpdateEventHandler Update;
```
with
```[csharp]
public delegate void SQLiteUpdateEventHandler(object sender, UpdateEventArgs e);

public class UpdateEventArgs : EventArgs
{
  public readonly string Database;
  public readonly string Table;
[...]

PR #400, opened 2017-08-18
Implement Data Change Notifications Callback

AlexanderTaeschner:
Adds an Update event.

Addresses #357.


src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -377,5 +384,30 @@ public virtual void EnableExtensions(bool enable = true)
...
+        /// Raises the Microsoft.Data.Sqlite.SqliteConnection.Update event.
+        /// </summary>
+        /// <param name="e">A Microsoft.Data.Sqlite.UpdateEventArgs that contains the event data.</param>
+        protected virtual void OnUpdate(UpdateEventArgs e)
+            => Update?.Invoke(this, e);
+
+        private void UpdateHook(object user_data, int type, string database, string table, long rowid)
+            => OnUpdate(new UpdateEventArgs(ConvertToUpdateEventType(type), database, table, rowid));
+
+        private UpdateEventType ConvertToUpdateEventType(int type)
```
bricelam: Probably worth using the same underlying values in the enum so we can just cast (like `SqliteType`). I think this would also remove the need for `Unknown`.

src/Microsoft.Data.Sqlite.Core/UpdateEventArgs.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <value>
+        /// The rowid of the affected row.
+        /// </value>
+        public long RowId { get; }
+    }
+
+    /// <summary>
+    /// Represents the event that changed the table data.
+    /// </summary>
+    public enum UpdateEventType
```
bricelam: Move to its own file.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: Done + eliminated UpdateHook method and used lambda instead.

src/Microsoft.Data.Sqlite.Core/UpdateEventArgs.cs
AlexanderTaeschner: Done.

AlexanderTaeschner:
I am not sure if we should include somehow the warning from https://sqlite.org/c3ref/update_hook.html

>The update hook implementation must not do anything that will modify the database connection that invoked the update hook. Any actions to modify the database connection must be deferred until after the completion of the sqlite3_step() call that triggered the update hook. Note that sqlite3_prepare_v2() and sqlite3_step() both modify their database connections for the meaning of "modify" in this paragraph.

Which basically means that no database queries can be issued inside the event handler.

Issue #405, opened 2017-08-23
Allow reading char from TEXT


bricelam:
Currently we call `GetInt64()` and cast it to `char`, but if the original value is `TEXT`, we should return the first character. We should als decide what happens when it's an empty string. (cast exception?)

bricelam:
We'll play it safe and only do this for single-character strings. All other values will use the existing code path.

PR #407, opened 2017-08-24
Allow reading char from TEXT

AlexanderTaeschner:
Addresses #405

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteValueReader.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteValueReader.cs
@@ -25,7 +25,19 @@ namespace Microsoft.Data.Sqlite
             => (byte)GetInt64(ordinal);
 
         public virtual char GetChar(int ordinal)
-            => (char)GetInt64(ordinal);
+        {
+            var sqliteType = GetSqliteType(ordinal);
+            if (sqliteType == raw.SQLITE_TEXT)
+            {
+                var val = GetString(ordinal);
+                if (val.Length == 1)
+                {
+                    return val[0];
+                }
+            }
+
+            return (char)GetInt64(ordinal);
+        }
 
         public virtual DateTime GetDateTime(int ordinal)
         {
--- a/test/Microsoft.Data.Sqlite.Tests/SqliteDataReaderTest.cs
+++ b/test/Microsoft.Data.Sqlite.Tests/SqliteDataReaderTest.cs
@@ -82,6 +82,13 @@ namespace Microsoft.Data.Sqlite
                 (char)1);
 
         [Fact]
+        public void GetChar_works_with_text()
+            => GetX_works(
+                "SELECT 'A';",
+                r => r.GetChar(0),
+                'A');        
+
+        [Fact]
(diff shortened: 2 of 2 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -373,7 +373,23 @@ public override byte GetByte(int ordinal)
...
-            => (char)GetInt64(ordinal);
+        {
+            var sqliteType = GetSqliteType(ordinal);
+            switch (sqliteType)
+            {
+                case raw.SQLITE_FLOAT:
+                case raw.SQLITE_INTEGER:
+                    return (char)GetInt64(ordinal);
+                default:
+                    var val = GetString(ordinal);
```
bricelam: I think we should flip this. Only call `GetString()` if it's `SQLITE_TEXT` otherwise default to the current behavior.

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -373,7 +373,23 @@ public override byte GetByte(int ordinal)
...
+                    if (!string.IsNullOrEmpty(val))
```
bricelam: I wonder if this should be even stronger and only try if it's a single character long.

src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -373,7 +373,23 @@ public override byte GetByte(int ordinal)
...
+                    {
+                        return val[0];
+                    }
+
+                    throw new InvalidCastException();
```
bricelam: Might be better just to revert to the current behavior in this case...

bricelam:
We may not be able to do this since it might be considered a breaking change (although not really since the value before was just '\0'). I'll discuss with the team whether we should do it and what the behavior should be.

AlexanderTaeschner:
The newest version defaults back to the old behaviour except in case of a single text character,

bricelam:
Could you rebase this now that #355 is merged?

AlexanderTaeschner:
Okay, rebased but not tested locally.

src/Microsoft.Data.Sqlite.Core/SqliteValueReader.cs
```diff
@@ -25,7 +25,19 @@ public virtual byte GetByte(int ordinal)
             => (byte)GetInt64(ordinal);
 
         public virtual char GetChar(int ordinal)
-            => (char)GetInt64(ordinal);
+        {
+            var sqliteType = GetSqliteType(ordinal);
+            if (sqliteType == raw.SQLITE_TEXT)
+            {
+                var val = GetString(ordinal);
+                if (!string.IsNullOrEmpty(val) && (val.Length == 1))
```
bricelam: After #414, `val` can't be `null`.

AlexanderTaeschner: Simplified the test in newest commit.

PR #410, opened 2017-08-25
Use PackageLineup to manage PackageReference versions

natemcmaster:
Set PackageReference version from lineups.


Directory.Build.targets
```diff
@@ -1,6 +1,14 @@
...
+    <PropertyGroup>
+      <_BootstrapperFile Condition=" $([MSBuild]::IsOSUnixLike()) ">build.sh</_BootstrapperFile>
+      <_BootstrapperFile Condition="! $([MSBuild]::IsOSUnixLike()) ">build.cmd</_BootstrapperFile>
+      <_BootstrapperError>
+        Package references have not been pinned. Run './$(_BootstrapperFile) /t:Pin'.
+        Also, you can run './$(_BootstrapperFile) /t:Restore' which will pin *and* restore packages. '$(_BootstrapperFile)' can be found in '$(MSBuildThisFileDirectory)'.
+      </_BootstrapperError>
+    </PropertyGroup>
+
+    <Error Code="KRB1001" Text="$(_BootstrapperError.Trim())" />
```
bricelam: 😭😢😿

For the love of contributors, we must keep the following scenario working.

1. Clone
2. Open solution in Visual Studio
3. Build

natemcmaster: On a scale from 1 to 10, how important is this to you? When we discussed this in the design meeting with the runtime-eng team, we decided (though regrettable), that this requirement ok. Since we require contributors to ensure `build.cmd` passes anyways, we figured there wasn't much additional burden on contributors or devs to run this warmup command.

bricelam: Hmm, it seems work. How does it know what version of `Internal.AspNetCore.Sdk` to restore?

natemcmaster: > Hmm, it seems work. 

Even after `git clean -xfd`?

> How does it know what version of Internal.AspNetCore.Sdk to restore?

Same as the other packages: the "lineup" determines this package version. Internal.AspNetCore.Sdk specifically comes from the version bundled in KoreBuild. https://github.com/aspnet/BuildTools/pull/367

bricelam: > Even after git clean -xfd?

Yep. Can't figure out how. 😄

natemcmaster: It may be an in-memory caching issue with the project system. When I `git clean -xfd` my branch on open a new VS instance, this is what I see:

image
image

bricelam: Doh, I didn't have the commit in the branch.

bricelam: I'm still very much against this since IMHO it will drive contributors away, but the error is clear, so it's only a single paper cut.

natemcmaster: By the way, I'm hoping this requirement won't be around forever, but we need a few improvements to VS and NuGet first. They're currently backlogged. https://github.com/NuGet/Home/issues/4781

natemcmaster: > I'm still very much against this

It's not my favorite either. I'm definitely open to more feedback on this. Any ideas for how to make this action automatic on VS open? I looked for extensibility points, but couldn't find anything reliable. $(SolutionName).sln.{before,after}.targets didn't work because VS has it's own build manager, and there isn't a pre-restore hook that I could find.

bricelam:
Ugg. This forces you to use .NET Core 2.1.0. I can't sign off on this. Default user experience (latest stable release of VS) will give you these awful, misleading errors.

> The current .NET SDK does not support targeting .NET Core 2.0.  Either target .NET Core 1.1 or lower, or use a version of the .NET SDK that supports .NET Core 2.0.

>The current .NET SDK does not support targeting .NET Standard 2.0.  Either target .NET Standard 1.6 or lower, or use a version of the .NET SDK that supports .NET Standard 2.0.

>The version of Microsoft.NET.Sdk used by this project is insufficient to support references to libraries targeting .NET Standard 1.5 or higher.  Please install version 2.0 or higher of the .NET Core SDK.

natemcmaster:
Do you have .NET Core 2.0.0 SDK installed and on your PATH? This is a requirement for .NET Core/Standard 2.0, regardless of this change.

bricelam:
Yep. Vanilla install of 15.3 + .NET Core SDK 2.0

natemcmaster:
I don't know how my change would lead to this then:

> The current .NET SDK does not support targeting .NET Core 2.0. Either target .NET Core 1.1 or lower, or use a version of the .NET SDK that supports .NET Core 2.0.

I'm not changing TFMs.

bricelam:
I believe it's because of the copy of NuGet that ships inside VS. Installing (the EXE/MSI) the .NET Core SDK 2**.1** fixes the issue.

bricelam:
Basically, it can't find SDK 2.1.0 so it falls back to 1.1 (not 2.0).

natemcmaster:
@bricelam sounds like PATH issues. When you run `build.cmd /t:noop`, do you see warnings such as this?


image

bricelam:
We're required to remove it? 😞 I hate dogfooding.

natemcmaster:
Ok, I won't merge this yet, but I'm going to leave this PR open and move discussion to internal email thread so we can include the managers.

bricelam:
P.S. I'm OK doing this on 2.0.x patch branches. The workarounds are easy, just impossible to figure out.
```cmd
build /t:Restore
DEL global.json
```

smitpatel:
>  The workarounds are easy, just impossible to figure out.
build /t:Restore
DEL global.json

I hit the same when we upgraded to 2.1.0
There is very thin line between dogfood version of product and impossible to use product. And often we have ended up on the wrong side of the line.

natemcmaster:
I added some of these rough edges to the README and opened https://github.com/aspnet/BuildTools/issues/386 so we can follow up on other issues with setting up a local dev environment correctly.

PR #414, opened 2017-08-31
UDFs: Throw when NULL specified for non-nullable parameters

bricelam:


src/Microsoft.Data.Sqlite.Core/SqliteValueReader.cs
```diff
@@ -62,11 +56,14 @@ public virtual DateTimeOffset GetDateTimeOffset(int ordinal)
         }
 
         public virtual decimal GetDecimal(int ordinal)
+            => decimal.Parse(GetString(ordinal), NumberStyles.Number | NumberStyles.AllowExponent, CultureInfo.InvariantCulture);
+
+        public virtual double GetDouble(int ordinal)
             => IsDBNull(ordinal)
-                ? GetNull<decimal>()
-                : decimal.Parse(GetString(ordinal), NumberStyles.Number | NumberStyles.AllowExponent, CultureInfo.InvariantCulture);
+                ? throw new InvalidCastException()
```
bricelam: Filed #419

ajcvickers:
@bricelam Is there an issue for this? If not, can you give a brief explanation of what this change is for? (Sorry if we have discussed this already; things have been kinda busy.)

bricelam:
This was discussed during the design meeting.

The current behavior is to coerce `NULL` into the requested type (e.g. you'll get `0` for `long`).

We decided that it is more helpful to the user if we throw when `NULL` is passed as an argument to a non-nullable parameter.

bricelam:
For example:

``` csharp
connection.CreateFunction("myfunc", (long value) => value);
```

``` sql
SELECT myfunc(NULL);
```

Before this would return `0`. After this PR it will throw.

src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
```diff
@@ -25,20 +25,14 @@ public SqliteDataRecord(sqlite3_stmt stmt)
...
-        public override long GetInt64(int ordinal)
-            => IsDBNull(ordinal)
-                ? throw new InvalidCastException()
-                : raw.sqlite3_column_int64(_stmt, ordinal);
-
-        public override string GetString(int ordinal)
-            => IsDBNull(ordinal)
-                ? throw new InvalidCastException()
-                : raw.sqlite3_column_text(_stmt, ordinal);
+        protected override double GetDoubleCore(int ordinal)
```
ajcvickers: Isn't this a breaking change? Or is it only used by code new in 2.0?

bricelam: New code (and real `internal`)

Issue #417, opened 2017-09-05
SqliteParameter: Truncate data to Size


bricelam:
In other providers, the `DbParameter.Size` property is used to limit the amount of data transmitted to the server. For example, strings and blobs longer than this are truncated before being sent to the server.

Microsoft.Data.Sqlite currently ignores this value.

PR #420, opened 2017-09-05
Implement SqliteParameter: Truncate data to Size

AlexanderTaeschner:
Addresses #417

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
@@ -144,7 +144,10 @@ namespace Microsoft.Data.Sqlite
         /// <value>The maximum size, in bytes, of the parameter.</value>
         public override int Size
         {
-            get => _size ?? SqliteParameterBinder.GetSize(_value);
+            get => _size ?? (
+                _value is string stringValue ? stringValue.Length :
+                _value is byte[] byteArray ? byteArray.Length :
+                0);
             set => _size = value;
         }
 
--- a/src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Data.Sqlite
     {
         private string _parameterName = string.Empty;
         private object _value;
+        private int? _size;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SqliteParameter" /> class.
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
```diff
@@ -141,7 +141,7 @@ public override string ParameterName
         /// Gets or sets the maximum size, in bytes, of the parameter.
         /// </summary>
         /// <value>The maximum size, in bytes, of the parameter.</value>
-        public override int Size { get; set; }
+        public override int Size { get; set; } = int.MaxValue;
```
bricelam: If you get the value before it's explicitly set, it should return the size of the `string` or `byte[]`. (Use an `int?` field to back it.) It can return `0` for other types.

AlexanderTaeschner: Okay, changed as suggested.

src/Microsoft.Data.Sqlite.Core/SqliteParameterBinder.cs
```diff
@@ -30,6 +42,29 @@ protected override void BindNull()
...
+            var text = value;
+            if (_size.HasValue && (value.Length > _size.Value))
+            {
+                text = value.Substring(0, _size.Value);
+            }
+
+            raw.sqlite3_bind_text(_stmt, _index, text);
+        }
+
+        internal static int GetSize(object value)
```
bricelam: I'd inline this logic with `SqliteParameter.Size` until something else needs it.

src/Microsoft.Data.Sqlite.Core/SqliteParameterBinder.cs
```diff
@@ -30,6 +42,29 @@ protected override void BindNull()
...
+        {
+            var type = value.GetType().UnwrapNullableType().UnwrapEnumType();
```
bricelam: `string` and `byte[]` can't be `Nullable` or `Enum`.

src/Microsoft.Data.Sqlite.Core/SqliteParameterBinder.cs
```diff
@@ -30,6 +42,29 @@ protected override void BindNull()
...
+            if (type == typeof(string))
+            {
+                return ((string)value).Length;
+            }
```
bricelam: Can simplify to
``` csharp
if (value is string stringValue)
{
    return stringValue.Length;
}
```

(same below)

AlexanderTaeschner:
Inlined the call to GetSize into  SqliteParameter.Size  and used pattern matching.

Issue #418, opened 2017-09-05
SqliteParamter: Honor SqliteType


bricelam:
By default, we bind non-primitive types using the most lossless and efficient format. There are, however cases where you want to use an alternative format. `SqliteDataReader` already supports additional formats. We should enable explicitly setting `SqliteParamter.SqliteType` to indicate an alternative format should be used:

Value | SqliteType | Description
--- | --- | ---
char | Text | A single-character string
DateTime | Real | A Julian day value
DateTimeOffset | Real | (Same as DateTime)
Guid | Text | A GUID string
TimeSpan | Real | The fractional days

PR #421, opened 2017-09-05
Implements SqliteParamter: Honor SqliteType

AlexanderTaeschner:
In order to leave the default behaviour (for char and Guid) unchanged I added SqliteType.Undefined and used it as new default value for SqliteType (was Text before).

Addresses #418

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Data.Sqlite
         private string _parameterName = string.Empty;
         private object _value;
         private int? _size;
+        private SqliteType? _sqliteType;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SqliteParameter" /> class.
@@ -96,7 +97,11 @@ namespace Microsoft.Data.Sqlite
         /// <value>The SQLite type of the parameter.</value>
         /// <remarks>Due to SQLite's dynamic type system, parameter values are not converted.</remarks>
         /// <seealso href="http://sqlite.org/datatype3.html">Datatypes In SQLite Version 3</seealso>
-        public virtual SqliteType SqliteType { get; set; } = SqliteType.Text;
+        public virtual SqliteType SqliteType
+        {
+            get => _sqliteType ?? SqliteValueBinder.GetSqliteType(_value);
+            set => _sqliteType = value;
+        }
 
         /// <summary>
         /// Gets or sets the direction of the parameter. Only <see cref="ParameterDirection.Input" /> is supported.
@@ -212,7 +217,7 @@ namespace Microsoft.Data.Sqlite
                 throw new InvalidOperationException(Resources.RequiresSet(nameof(Value)));
             }
 
-            new SqliteParameterBinder(stmt, index, _value, _size).Bind();
+            new SqliteParameterBinder(stmt, index, _value, _size, _sqliteType).Bind();
 
             return true;
         }
--- a/src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using Microsoft.Data.Sqlite.Properties;
 
@@ -11,10 +12,17 @@ namespace Microsoft.Data.Sqlite
     internal abstract class SqliteValueBinder
     {
         private readonly object _value;
+        private readonly SqliteType? _sqliteType;
 
         protected SqliteValueBinder(object value)
+            : this(value, null)
+        {
+        }
+
+        protected SqliteValueBinder(object value, SqliteType? sqliteType)
         {
             _value = value;
+            _sqliteType = sqliteType;
         }
 
         protected abstract void BindInt64(long value);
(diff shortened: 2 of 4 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
```diff
@@ -95,7 +95,7 @@ public SqliteParameter(string name, SqliteType type, int size, string sourceColu
         /// <value>The SQLite type of the parameter.</value>
         /// <remarks>Due to SQLite's dynamic type system, parameter values are not converted.</remarks>
         /// <seealso href="http://sqlite.org/datatype3.html">Datatypes In SQLite Version 3</seealso>
-        public virtual SqliteType SqliteType { get; set; } = SqliteType.Text;
+        public virtual SqliteType SqliteType { get; set; } = SqliteType.Undefined;
```
bricelam: Use a `SqliteType?` field instead. If they read the value without setting it, we should infer the type from `_value`.

src/Microsoft.Data.Sqlite.Core/SqliteType.cs
```diff
@@ -12,6 +12,11 @@ namespace Microsoft.Data.Sqlite
     public enum SqliteType
     {
         /// <summary>
+        /// Type affinity is not set.
+        /// </summary>
+        Undefined = -1,
```
bricelam: Represent with `null` instead.

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
```diff
@@ -128,8 +170,16 @@ public virtual void Bind()
             }
             else if (type == typeof(TimeSpan))
             {
-                var value = ((TimeSpan)_value).ToString("c");
-                BindText(value);
+                if (_sqliteType == SqliteType.Real)
+                {
+                    var value = ((TimeSpan)_value).Ticks / (double)TimeSpan.TicksPerDay;
```
bricelam: `.TotalDays`

AlexanderTaeschner:
Changed as suggested.

src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
```diff
@@ -95,7 +96,11 @@ public SqliteParameter(string name, SqliteType type, int size, string sourceColu
         /// <value>The SQLite type of the parameter.</value>
         /// <remarks>Due to SQLite's dynamic type system, parameter values are not converted.</remarks>
         /// <seealso href="http://sqlite.org/datatype3.html">Datatypes In SQLite Version 3</seealso>
-        public virtual SqliteType SqliteType { get; set; } = SqliteType.Text;
+        public virtual SqliteType SqliteType
+        {
+            get => _sqliteType.HasValue ? _sqliteType.Value : SqliteValueBinder.GetSqliteType(_value);
```
smitpatel: use null conditional & coalesce?

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
```diff
@@ -64,18 +71,44 @@ public virtual void Bind()
...
-                var value = (long)(char)_value;
-                BindInt64(value);
+                var chr = (char)_value;
+                if (_sqliteType == SqliteType.Text)
+                {
+                    BindText(new string(chr, 1));
+                }
+                else
+                {
+                    var value = (long)chr;
```
smitpatel: Should this value be inlined? If no then new string should be saved in value too.

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
```diff
@@ -151,5 +202,123 @@ public virtual void Bind()
...
         }
+
+        internal static SqliteType GetSqliteType(object value)
+        {
+            if (value == null)
+            {
+                return SqliteType.Text;
+            }
+
+            var type = value.GetType().UnwrapNullableType().UnwrapEnumType();
```
smitpatel: use switch on type and pattern matching of C# 7.0

smitpatel: is switch is not possible then dictionary may be

src/Microsoft.Data.Sqlite.Core/SqliteParameter.cs
AlexanderTaeschner: Done.

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
AlexanderTaeschner: Done.

test/Microsoft.Data.Sqlite.Tests/SqliteParameterTest.cs
```diff
@@ -165,13 +166,18 @@ public void Bind_is_noop_on_unknown_parameter()
...
         {
             using (var connection = new SqliteConnection("Data Source=:memory:"))
             {
                 var command = connection.CreateCommand();
                 command.CommandText = "SELECT @Parameter;";
-                command.Parameters.AddWithValue("@Parameter", value);
+                var sqliteParameter = command.Parameters.AddWithValue("@Parameter", value);
+                if (sqliteType.HasValue)
+                {
+                    sqliteParameter.SqliteType = sqliteType.Value;
```
smitpatel: does need check here. null will anyway fall back to default

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
```diff
@@ -64,18 +71,45 @@ public virtual void Bind()
             }
             else if (type == typeof(char))
             {
-                var value = (long)(char)_value;
-                BindInt64(value);
+                var chr = (char)_value;
+                if (_sqliteType == SqliteType.Text)
+                {
+                    var value = new string(chr, 1);
```
smitpatel: @bricelam - Are all this temp variables are for readability?

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
bricelam: +1 for using a dictionary

test/Microsoft.Data.Sqlite.Tests/SqliteParameterTest.cs
AlexanderTaeschner: sqliteParameter.SqliteType is not nullable and sqliteType.Value throws if sqliteType has no value.

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
AlexanderTaeschner: Changed to using a dictionary in the newest version.

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
bricelam: Copy-paste. I wanted to go through and clean it up a bit (And switch to  C# 7 `case` statements in a few places.)

smitpatel: @AlexanderTaeschner - You can leave these as is. We will clean it up later.
R# :trollface:

src/Microsoft.Data.Sqlite.Core/SqliteValueBinder.cs
```diff
@@ -98,8 +133,17 @@ public virtual void Bind()
             }
             else if (type == typeof(Guid))
             {
-                var value = ((Guid)_value).ToByteArray();
-                BindBlob(value);
+                var guid = (Guid)_value;
+                if (_sqliteType == SqliteType.Text)
+                {
+                    var value = guid.ToString();
```
bricelam: I wonder if we should call `.ToUpper()` on this...

bricelam: (Seems like the most canonical format to me.)

bricelam:
Note to self/merger: We probably need to react to this in the EF Core tests. (DbParameter logging baselines)

Issue #419, opened 2017-09-05
Don't throw InvalidCastException


bricelam:
SQL Client tries to throw the following when it encounters an unexpected `NULL` value.

> SqlNullValueException: Data is Null. This method or property cannot be called on Null values.

We throw the following which isn't very helpful.

> InvalidCastException: Specified cast is not valid.

PR #427, opened 2017-09-11
Don't throw InvalidCastException, throw customized InvalidOperationException instead

AlexanderTaeschner:
Addresses #419

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using Microsoft.Data.Sqlite.Properties;
 using SQLitePCL;
 
 namespace Microsoft.Data.Sqlite
@@ -24,6 +25,9 @@ namespace Microsoft.Data.Sqlite
 
         public override int FieldCount
             => raw.sqlite3_column_count(_stmt);
+
+        protected override string OnNullErrorMsg
+            => Resources.CalledOnNullValue;
 
         protected override double GetDoubleCore(int ordinal)
             => raw.sqlite3_column_double(_stmt, ordinal);
@@ -53,7 +57,7 @@ namespace Microsoft.Data.Sqlite
         protected override T GetNull<T>()
             => typeof(T) == typeof(DBNull) || typeof(T) == typeof(object)
                 ? (T)(object)DBNull.Value
-                : throw new InvalidCastException();
+                : throw new InvalidOperationException(Resources.CalledOnNullValue);
 
         public virtual string GetName(int ordinal)
         {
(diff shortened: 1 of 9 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
```diff
@@ -53,7 +54,7 @@ protected override int GetSqliteType(int ordinal)
         protected override T GetNull<T>()
             => typeof(T) == typeof(DBNull) || typeof(T) == typeof(object)
                 ? (T)(object)DBNull.Value
-                : throw new InvalidCastException();
+                : throw new SqlNullValueException();
```
bricelam: I think this type is SQL Client specific and we should introduce our own type. I'll look more into it...

bricelam: Confirmed.

> The System.Data.SqlTypes namespace provides classes for native data types in SQL Server.

I think we should just throw `InvlaidOperationException` with a message like this:

> The data is NULL. This method can't be called on NULL values. Check using IsDBNull before calling.

AlexanderTaeschner:
Updated patch with customized InvalidOperationException.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
```diff
@@ -538,7 +538,7 @@ public void CreateFunction_works_when_parameter_null_but_type_long()
                 var ex = Assert.Throws<SqliteException>(() => connection.ExecuteScalar<long>("SELECT test(NULL);"));
 
                 Assert.Equal(
-                    Resources.SqliteNativeError(raw.SQLITE_ERROR, new InvalidCastException().Message),
+                    Resources.SqliteNativeError(raw.SQLITE_ERROR, Resources.CalledOnNullValue),
```
bricelam: Hmm, the error message doesn't make much sense in this context. Can we customize it for SqliteParameterBinder?

AlexanderTaeschner: Okay, done in the newest revision.

AlexanderTaeschner:
Rebased on top of current dev.

src/Microsoft.Data.Sqlite.Core/SqliteValueReader.cs
```diff
@@ -117,24 +118,28 @@ public virtual int GetInt32(int ordinal)
...
-                ? throw new InvalidCastException()
+                ? throw new InvalidOperationException(Resources.CalledOnNullValue)
                 : GetStringCore(ordinal);
 
         protected abstract string GetStringCore(int ordinal);
 
         public virtual T GetFieldValue<T>(int ordinal)
+            => GetFieldValue<T>(ordinal, Resources.CalledOnNullValue);
+
+        public virtual T GetFieldValue<T>(int ordinal, string onNullErrorMsg)
```
bricelam: Callers should have to pass in the message. Can we simply override it (via an protected abstract property or something) in `SqliteDataRecord` and `SqliteParamterReader`? You can flow the function name in through the `SqliteParameterReader` constructor.

AlexanderTaeschner:
Rebased on current dev and added protected abstract property with error message to SqliteValueReader.

Issue #436, opened 2017-10-02
Improve SqliteDataReader.GetSchemaTable() DataType


bricelam:
The `DataType` returned by `SqliteDataReader.GetSchemaTable()` has a couple problems:
1. If the first row contains `NULL`, it defaults to `int` (see #435)
2. If the column contains mixed types, the type of the first row is used.

I propose we fix this by sampling the rows for the most used type. For example, by using a query like this:

``` sql
SELECT typeof(column1)
FROM table1
WHERE typeof(column1) != 'NULL'
GROUP BY typeof(column1)
ORDER BY count() DESC
LIMIT 1;
```

AlexanderTaeschner:
I am not sure if this query is really necessary. The following test already succeeds without any code changes:
```csharp
        [Fact]
        public void GetSchemaTable_DataType_works()
        {
            using (var connection = new SqliteConnection("Data Source=:memory:"))
            {
                connection.Open();
                connection.ExecuteNonQuery(
                    "CREATE TABLE Test (Name NVARCHAR(100));");
                connection.ExecuteNonQuery("INSERT INTO Test VALUES(1);");

                using (var reader = connection.ExecuteReader("SELECT Name FROM Test;"))
                {
                    var schema = reader.GetSchemaTable();
                    Assert.True(schema.Columns.Contains("DataType"));
                    Assert.True(schema.Columns.Contains("DataTypeName"));

                    Assert.Equal(1, schema.Rows.Count);
                    
                    Assert.Equal(typeof(String), schema.Rows[0]["DataType"]);
                    Assert.Equal("NVARCHAR", schema.Rows[0]["DataTypeName"]);
                }
            }
        }
```

bricelam:
The case is more like this:
``` sql
CREATE TABLE test(value);
INSERT INTO test(value) VALUES (NULL), ('A'), (1);
```

AlexanderTaeschner:
Your proposed query
`SELECT typeof(value) FROM test WHERE typeof(value) != 'null' GROUP BY typeof(value) ORDER BY count() DESC LIMIT 1;`
would lead to `integer` for your example, but with this type neither the `NULL` nor the `'A'` value could be read out. The only compatible datatype would be `text` or as .NET type `String`.

bricelam:
Yes, bad example, but you get a better idea of the problem.

PR #437, opened 2017-10-04
Improve SqliteDataReader.GetSchemaTable() DataType

AlexanderTaeschner:
Note that column and table names cannot be bound using parameters (I enclosed them in " to make SQL injection attacks less likely).

Addresses #436

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
@@ -540,6 +540,19 @@ namespace Microsoft.Data.Sqlite
 
                         var cnt = (long)command.ExecuteScalar();
                         schemaRow[IsUnique] = cnt != 0;
+
+                        command.Parameters.Clear();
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteDataReader.cs
```diff
@@ -538,6 +538,18 @@ public override DataTable GetSchemaTable()
 
                         var cnt = (long)command.ExecuteScalar();
                         schemaRow[IsUnique] = cnt != 0;
+
+                        command.Parameters.Clear();
+                        command.CommandText = new StringBuilder()
+                            .AppendLine($"SELECT typeof(\"{columnName}\")")
```
bricelam: Looks like column names can contain quotes... Do a `.Replace("\"", "\"\"")` just in case.

AlexanderTaeschner: Done.

AlexanderTaeschner:
Rebased on top of current dev.

PR #442, opened 2017-10-09
Run Code Cleanup

smitpatel:


src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateAggregate.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Data.Sqlite
     partial class SqliteConnection
     {
         /// <summary>
-        /// Creates or redefines an aggregate SQL function.
+        ///     Creates or redefines an aggregate SQL function.
```
bricelam: Change the .tt file

src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateFunction.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Data.Sqlite
     partial class SqliteConnection
     {
         /// <summary>
-        /// Creates or redfines a SQL function.
+        ///     Creates or redfines a SQL function.
```
bricelam: Change the .tt file

src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateAggregate.cs
smitpatel: Updated.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.CreateFunction.cs
smitpatel: Updated.

PR #447, opened 2017-10-16
Add RepositoryRoot

ryanbrandenburg:


bricelam:
Can't we derive this in KoreBuild?

E.g. `$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.cmd))`

ryanbrandenburg:
KoreBuild does derive this, but we run into issue like https://github.com/aspnet/BuildTools/issues/427 when  running in VS because VS doesn't run KoreBuild.

ryanbrandenburg:
@bricelam does that address your concerns?

Issue #443, opened 2017-10-10
Implement DbConnection.GetSchema


Thorium:
Is there a way to get basic meta data information from the database,?

I would actually need just need initial version of System.Data.Common style `GetSchema` supporting these:

- DataTypes: 
   - DataType, TypeName, ProviderDbType
   - Gives the information how this provider maps types to Sqlite database
- Tables
   - TABLE_TYPE, TABLE_CATALOG, TABLE_NAME
   - List of tables in the database, type is only interesting not to be a SYSTEM_TABLE.
- ForeignKeys
   - CONSTRAINT_NAME, TABLE_CATALOG, TABLE_NAME, FKEY_TO_CATALOG, FKEY_TO_TABLE, FKEY_TO_COLUMN, FKEY_FROM_COLUMN 
   - Information of relations between tables.

That would be enough, as table column names and primary key info can be gotten from `pragma table_info`.

Thorium:
Data types are constant:
```
DataType, TypeName, ProviderDbType
'System.Int16','smallint','10'
'System.Int32','int','11'
'System.Double','real','8'
'System.Single','single','15'
'System.Double','float','8'
'System.Double','double','8'
'System.Decimal','money','7'
'System.Decimal','currency','7'
'System.Decimal','decimal','7'
'System.Decimal','numeric','7'
'System.Boolean','bit','3'
'System.Boolean','yesno','3'
'System.Boolean','logical','3'
'System.Boolean','bool','3'
'System.Boolean','boolean','3'
'System.Byte','tinyint','2'
'System.Int64','integer','12'
'System.Int64','counter','12'
'System.Int64','autoincrement','12'
'System.Int64','identity','12'
'System.Int64','long','12'
'System.Int64','bigint','12'
'System.Byte[]','binary','1'
'System.Byte[]','varbinary','1'
'System.Byte[]','blob','1'
'System.Byte[]','image','1'
'System.Byte[]','general','1'
'System.Byte[]','oleobject','1'
'System.String','varchar','16'
'System.String','nvarchar','16'
'System.String','memo','16'
'System.String','longtext','16'
'System.String','note','16'
'System.String','text','16'
'System.String','ntext','16'
'System.String','string','16'
'System.String','char','16'
'System.String','nchar','16'
'System.DateTime','datetime','6'
'System.DateTime','smalldate','6'
'System.DateTime','timestamp','6'
'System.DateTime','date','6'
'System.DateTime','time','6'
[...]

PR #449, opened 2017-10-18
Initial implementation for DBConnection.GetSchema(String) #443

Thorium:


Thorium:
The Appveyor test fails as I was naive and used the test database file `readonly.db` that was used by the other tests. However I tested locally with larger database and it works. How should the databases be used in the unit tests?

bricelam:
> How should the databases be used in the unit tests?

Use `:memory:` unless the test requires otherwise.

test/Microsoft.Data.Sqlite.Tests/SqliteConnectionTest.cs
```diff
@@ -890,5 +890,26 @@ public void DataChange_event_works()
                 Assert.Equal(101, list[0].RowId);
             }
         }
+
+        [Fact]
+        public void GetSchema_has_collections()
+        {
+            var connectionString = "Data Source=readonly.db";
```
smitpatel: try datasource as `:memory:` ?

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -556,5 +556,138 @@ public AggregateContext(T seed)
...
+        /// Returns schema information for the data source of this DbConnection using the specified string for the schema name.
+        /// </summary>
+        /// <param name="collectionName">Specifies the name of the schema to return.</param>
+        /// <returns>A DataTable that contains schema information.</returns>
+        public override System.Data.DataTable GetSchema(string collectionName)
+        {
+            var dt = new DataTable(collectionName);
+            switch (collectionName)
+            {
+                case "DataTypes":
```
bricelam: Use `DbMetaDataCollectionNames.DataTypes`

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -556,5 +556,138 @@ public AggregateContext(T seed)
...
+                    dt.Columns.AddRange(new[] {
+                        new DataColumn("DataType", typeof(string)),
+                        new DataColumn("TypeName", typeof(string)),
+                        new DataColumn("ProviderDbType", typeof(int))
```
bricelam: Use `DbMetaDataColumnNames.DataType`, etc.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -556,5 +556,138 @@ public AggregateContext(T seed)
...
+                        new DataColumn("TABLE_NAME"),
+                        new DataColumn("FKEY_TO_CATALOG"),
+                        new DataColumn("TABLE_CATALOG"),
+                        new DataColumn("FKEY_TO_TABLE"),
+                        new DataColumn("FKEY_FROM_COLUMN"),
+                        new DataColumn("FKEY_TO_COLUMN"),
+                        new DataColumn("CONSTRAINT_NAME")
+                    });
+                    foreach(var tablename in tables){
+                        var relationQuery = "pragma foreign_key_list(" + tablename + ")";
```
bricelam: Can combine this and the previous query using this syntax:
``` sql
SELECT t.name,
       fk.table,
       fk.from,
       fk.to
FROM sqlite_master t,
     pragma_foreign_key_list(t.name) fk
WHERE t.type = 'table'
```

bricelam:
We should probably implement the *MetaDataCollections* collection to show which collections we currently support.

For the collections we do support, we should probably include all the "usual" columns (even if we don't populate them with data).

bricelam:
Let's add a few more tests to make sure the data returned is correct.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
Thorium: Nice idea, however [DB Browser for SQLite](http://sqlitebrowser.org/) doesn't support that, so I don't know how widely that is supported.

Thorium:
@bricelam MetaDataCollections and more testing added.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -556,5 +556,162 @@ public AggregateContext(T seed)
...
+
+        /// <summary>
+        /// System.Data.Common, initial implementation of API DBConnection.GetSchema(String)
+        /// Returns schema information for the data source of this DbConnection using the specified string for the schema name.
+        /// </summary>
+        /// <param name="collectionName">Specifies the name of the schema to return.</param>
+        /// <returns>A DataTable that contains schema information.</returns>
+        public override System.Data.DataTable GetSchema(string collectionName)
+        {
+            return GetSchema(collectionName, null);
```
bricelam: I suppose we should also add:
``` csharp
public override DataTable GetSchema()
    => GetSchema(DbMetaDataCollectionNames.MetaDataCollections, null);
```

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -556,5 +556,162 @@ public AggregateContext(T seed)
...
+                    dt.Rows.Add(new object[] { "System.String","string",16 });
+                    dt.Rows.Add(new object[] { "System.String","char",16 });
+                    dt.Rows.Add(new object[] { "System.String","nchar",16 });
+                    dt.Rows.Add(new object[] { "System.DateTime","datetime",6 });
+                    dt.Rows.Add(new object[] { "System.DateTime","smalldate",6 });
+                    dt.Rows.Add(new object[] { "System.DateTime","timestamp",6 });
+                    dt.Rows.Add(new object[] { "System.DateTime","date",6 });
+                    dt.Rows.Add(new object[] { "System.DateTime","time",6 });
+                    dt.Rows.Add(new object[] { "System.Guid","uniqueidentifier",4 });
+                    dt.Rows.Add(new object[] { "System.Guid","guid",4 });
```
bricelam: We need to refine this information before merging.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
Thorium: done

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
Thorium: The information here describes what types of .NET can be mapped to types that will be understood by the database driver, and the corresponding System.Data.Common DbTypes. This information is taken from the System.Data.SQLite.dll library.

I think this is more design decision. Do you want to continue from here?
I'm already happy with the current implementation but feel free to change what ever you need.

bricelam:
Closing this per [our discussion](https://github.com/aspnet/Microsoft.Data.Sqlite/issues/443#issuecomment-339444178). Querying directly for this metadata (and hard-coding a list of supported types) is the correct way to handle this for the time being.

Issue #404, opened 2017-08-23
Add extended error code to SqliteException


evil-shrike:
Currently (2.0) we have only `int` field `SqliteException.SqliteErrorCode` with an error code.

For example for a constraint violation it'll be 19. And `SqliteException.Message` will be `SQLite Error 19: 'UNIQUE constraint failed: User.Login'`.

Constraint violation is a very broad case. It can be unique constrain, check, foreign key and so on. This information is absent in SqliteException. Also `Message` property value is a combination "`SQLite Error N:" and original text from SQLite - 'UNIQUE constraint failed: User.Login'. This is bad for parsing. 

What's needed:

* Original text from SQLite: ('UNIQUE constraint failed: User.Login')
* SqliteErrorCode should be a enumration
* there should be an additional error code with mor details.

Here's how it's implemented in System.Data.SQLite. Connection object provider `ExtendedResultCode` method which return more detailed error code (using the same enumration `SQLiteErrorCode` for general code and details code is bad I think).
 
```
SQLiteException originalEx = ..;
SQLiteErrorCode code = m_connection.ExtendedResultCode();			
if (originalEx.ResultCode == SQLiteErrorCode.Constraint) {
  if (code == SQLiteErrorCode.Constraint_ForeignKey) 
  {
    // foreign key violation
  }
  else if (code == SQLiteErrorCode.Constraint_PrimaryKey) 
  {
    // primary key violation
  }
}
```

bricelam:
SQLite has a limited set of [error codes](https://www.sqlite.org/rescode.html).

You can turn on extended error codes by dropping down to SQLitePCL.raw:

```csharp
SQLitePCL.raw.sqlite3_extended_result_codes(connection.Handle, 1);
```

> Original text from SQLite

I don't think this is useful after turning on extended error codes.

> SqliteErrorCode should be a enumration

The application or custom builds of SQLite can define additional error codes, so the set is not finite. You can always use the constants defined in SQLitePCL.raw:

```csharp
ex.SqliteErrorCode == SQLitePCL.raw.SQLITE_CONSTRAINT_FOREIGNKEY
```

> there should be an additional error code with mor details

Agreed. We should expose the extended error code on `SqliteException` without explicitly having to turn them on. (We can do this using `sqlite3_extended_errcode()`.)

evil-shrike:
>I don't think this is useful after turning on extended error codes. Also, exceptions are supposed to be localized so parsing them should be avoided.

[...]

PR #451, opened 2017-10-27
Use extended Sqlite error codes

AlexanderTaeschner:
Addresses #404

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteException.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteException.cs
@@ -19,8 +19,21 @@ namespace Microsoft.Data.Sqlite
         /// <param name="message">The message to display for the exception. Can be null.</param>
         /// <param name="errorCode">The SQLite error code.</param>
         public SqliteException(string message, int errorCode)
+            : this(message, errorCode, errorCode)
+        { }
+
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="SqliteException" /> class.
+        /// </summary>
+        /// <param name="message">The message to display for the exception. Can be null.</param>
+        /// <param name="errorCode">The SQLite error code.</param>
+        /// /// <param name="extendedErrorCode">The extended SQLite error code.</param>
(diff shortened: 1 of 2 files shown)


bricelam:
This PR raises another concern: If the app calls `sqlite3_extended_result_codes()` does it break us? If so, we should file a new issue to be resilient to this.

AlexanderTaeschner:
Changed the PR by adding the property to SqliteException.

If someone calls ` sqlite3_extended_result_codes()` it breaks the check https://github.com/aspnet/Microsoft.Data.Sqlite/blob/791993269cfbf97ec3fbdb5139d860ba61ef7396/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs#L355
we could add `|| rc == raw.SQLITE_LOCKED_SHAREDCACHE` to be safe.

bricelam:
Yeah, seems like we should update the check

src/Microsoft.Data.Sqlite.Core/SqliteException.cs
```diff
@@ -49,8 +69,11 @@ public static void ThrowExceptionForRC(int rc, sqlite3 db)
             var message = db == null || db.ptr == IntPtr.Zero
                 ? raw.sqlite3_errstr(rc) + " " + Resources.DefaultNativeError
                 : raw.sqlite3_errmsg(db);
+            var extendedErrorCode = db == null || db.ptr == IntPtr.Zero
+                ? rc
+                : raw.sqlite3_extended_errcode(db);
```
bricelam: DRY:
``` csharp
string message;
int extendedErrorCode;
if (db == null || db.ptr == IntPtr.Zero)
{
    extendedErrorCode = rc;
    message = raw.sqlite3_errstr(rc) + " " + Resources.DefaultNativeError;
}
else
{
    message  = raw.sqlite3_errmsg(db);
    extendedErrorCode  = raw.sqlite3_extended_errcode(db);
}
```

AlexanderTaeschner: Rebased on top of dev and corrected the DRY issue.

PR #452, opened 2017-10-31
Pin tool and package versions to make builds more repeatable

natemcmaster:
Part of aspnet/Universe#575 and aspnet/KoreBuild#239

This removes floating versions of dependencies. This includes the korebuild version and versions used by PackageReference's. It also removes 'PackageLineup' which means contributors and devs aren't required to run 'build.cmd /t:Pin' anymore to use a clean checkout in VS.


natemcmaster:
:bell:

PR #456, opened 2017-11-10
Target netcoreapp2.1 in tests and samples

pranavkm:


samples/AggregateFunctionSample/AggregateFunctionSample.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
pranavkm: @bricelam should these samples cross-compile for netcoreapp2.1 and netcoreapp2.0? Didn't seem like there was value in doing so since they aren't tested as part of a test.

bricelam: Just 2.1 seems fine.

Issue #17, opened 2014-04-02
Online backup


bricelam:
See [Online Backup API](http://sqlite.org/c3ref/backup_finish.html).

zsims:
I'm interested in looking at this in the near future, do you have any preferences for how the API would be exposed?

I'm thinking to extend `SqliteConnection` with:
```
void Backup(string sourceDatabaseName, SqliteConnection destination, string dedtinationDatabaseName);
```

Which would do the backup all pages. Progress or incremental page backups could come later if required. Thoughts?

ErikEJ:
System.Data.SQLite already has BackupDatabase https://github.com/OpenDataSpace/System.Data.SQLite/blob/master/System.Data.SQLite/SQLiteConnection.cs#L792 (not official source code!)

bricelam:
👍 In general, we've tried to keep our APIs compatible with System.Data.SQLite unless there is a compelling reason to change them

bricelam:
lol well, maybe that statement isn't true... but it makes sense to do here. And we did it for `EnableExtensions`.

HardRock88:
Any chance to get this feature working any time soon ?

bricelam:
It would make a great contribution! Unfortunately, our team is currently focused on higher priority items and won't have the bandwidth to do this any time soon.

AlexanderTaeschner:
System.Data.SQLite has the following interface:
```csharp
public void BackupDatabase( 
    SQLiteConnection destination, 
    string destinationName, 
    string sourceName, 
    int pages, 
    SQLiteBackupCallback callback, 
    int retryMilliseconds 
)
```
I have two questions:
- Do we need the callback hook (at the moment I see no real use case for it)?
- Should we add default values (eg. sourceName = "main", pages = -1 (full backup))?

bricelam:
Hmm, that's pretty far removed from the native API. I'd like to investigate some alternative designs. Prehaps using some modern .NET async patterns...

AlexanderTaeschner:
I would propose to add three overloads to SQLiteConnection:
```csharp
public virtual void BackupDatabase(SqliteConnection destination)
    => BackupDatabase(destination, "main", "main");
public virtual void BackupDatabase(SqliteConnection destination, string destinationName, string sourceName)
    => BackupDatabase(destination, destinationName, sourceName, -1);
public virtual void BackupDatabase(SqliteConnection destination, string destinationName, string sourceName, int pages)
```
[...]

PR #463, opened 2017-12-05
Add online backup.

AlexanderTaeschner:
Addresses #17

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
@@ -380,6 +380,41 @@ namespace Microsoft.Data.Sqlite
             SqliteException.ThrowExceptionForRC(rc, _db);
         }
 
+        /// <summary>
+        ///     Backup of the connected database.
+        /// </summary>
+        /// <param name="destination">The destination of the backup.</param>
+        public virtual void BackupDatabase(SqliteConnection destination)
+            => BackupDatabase(destination, "main", "main");
+
+        /// <summary>
+        ///     Backup of the connected database.
+        /// </summary>
+        /// <param name="destination">The destination of the backup.</param>
+        /// <param name="destinationName">The name of the destination database.</param>
+        /// <param name="sourceName">The name of the source database.</param>
+        public virtual void BackupDatabase(SqliteConnection destination, string destinationName, string sourceName)
+        {
+            if (_db == null
+                || _db.ptr == IntPtr.Zero
+                || destination.Handle == null
+                || destination.Handle.ptr == IntPtr.Zero)
+            {
+                throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(BackupDatabase)));
+            }
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -380,6 +380,42 @@ public virtual void EnableExtensions(bool enable = true)
...
+            if (backup.ptr == IntPtr.Zero)
+            {
+                SqliteException.ThrowExceptionForDatabaseError(_db);
+            }
+
+            var rc = raw.sqlite3_backup_step(backup, -1);
+            SqliteException.ThrowExceptionForRC(rc, _db);
+
+            rc = raw.sqlite3_backup_finish(backup);
+            SqliteException.ThrowExceptionForRC(rc, _db);
```
bricelam: No need to call `sqlite3_backup_finish`, just wrap `backup` inside a `using` statement.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -380,6 +380,42 @@ public virtual void EnableExtensions(bool enable = true)
...
+                || destination.Handle == null
+                || destination.Handle.ptr == IntPtr.Zero)
+            {
+                throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(BackupDatabase)));
+            }
+
+            var backup = raw.sqlite3_backup_init(destination.Handle, destinationName, _db, sourceName);
+            if (backup.ptr == IntPtr.Zero)
+            {
+                SqliteException.ThrowExceptionForDatabaseError(_db);
```
bricelam: I wonder if we should be doing this elsewhere...

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
AlexanderTaeschner: Okay, changed.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -380,6 +380,41 @@ public virtual void EnableExtensions(bool enable = true)
...
+
+            using (var backup = raw.sqlite3_backup_init(destination.Handle, destinationName, _db, sourceName))
+            {
+                if (backup.ptr == IntPtr.Zero)
+                {
+                    SqliteException.ThrowExceptionForDatabaseError(_db);
+                }
+
+                var rc = raw.sqlite3_backup_step(backup, -1);
+                SqliteException.ThrowExceptionForRC(rc, _db);
```
bricelam: The documentation doesn't say, but it looks like (after a lot of digging) this is supposed to be `destincation._db` too. I'll fix.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
bricelam: (I checked and couldn't find any.)

PR #468, opened 2017-12-18
Override DbConnection.DbProviderFactory

### LINKED ISSUE #467: Override DbConnection.DbProviderFactory
We should start returning the singleton provider factory for Sqlite so that `DbProviderFactories.GetFactory(DbConnection)` starts to work in 2.1. 

cc @roji @alexkeh @bgrainger since this is probably missing in many ADO.NET providers that target .NET Standard 2.0.

bricelam:
Resolves #467

cc @divega


smitpatel:
Why the test is conditional on ~netcoreapp2.0?

bricelam:
`DbProviderFactories` was introduced in 2.1

PR #475, opened 2018-01-09
[2.0.1] Revert to using IMMEDIATE with Serializable transactions

### LINKED ISSUE #474: Deadlock (eventually times out) inside Serializable transactions
Here is some over-simplified code demonstrating the problem.

``` csharp
var connectionString = "Data Source=transations;Mode=Memory;Cache=Shared";
using (var connection1 = new SqliteConnection(connectionString))
using (var connection2 = new SqliteConnection(connectionString))
{
    connection1.Open();
    connection2.Open();

    connection1.ExecuteNonQuery("CREATE TABLE Data (Value); INSERT INTO Data VALUES (0);");

    using (connection1.BeginTransaction())
    using (connection2.BeginTransaction())
    {
        connection1.ExecuteNonQuery("SELECT * FROM Data");
        connection2.ExecuteNonQuery("SELECT * FROM Data");

        // Each command is waiting for the other transaction to complete
        connection1.ExecuteNonQuery("UPDATE Data SET Value = 1");
        connection2.ExecuteNonQuery("UPDATE Data SET Value = 2");
    }
}
```

In version 1.x, we avoided the deadlock since `connection2` couldn't begin the transaction until the one on `connection1` completed.

In version 2.0.0, we allow both transactions to begin which eventually leads to the deadlock. Note, one of the commands eventually times out and throws rolling back its transaction unblocking the application.

bricelam:
Fixes #474

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteTransaction.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteTransaction.cs
@@ -43,9 +43,10 @@ namespace Microsoft.Data.Sqlite
                 throw new ArgumentException(Resources.InvalidIsolationLevel(isolationLevel));
             }
 
-            // TODO: Register transaction hooks to detect when a user manually completes a transaction created using
-            //       this API
-            connection.ExecuteNonQuery("BEGIN;");
+            connection.ExecuteNonQuery(
+                IsolationLevel == IsolationLevel.Serializable
+                    ? "BEGIN IMMEDIATE;"
+                    : "BEGIN;");
         }
 
         /// <summary>
--- a/version.props
+++ b/version.props
@@ -1,6 +1,6 @@
 <!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
 <Project>
     <PropertyGroup>
-        <VersionPrefix>2.0.0</VersionPrefix>
+        <VersionPrefix>2.0.1</VersionPrefix>
     </PropertyGroup>
 </Project>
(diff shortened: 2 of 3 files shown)


version.props
```diff
@@ -1,6 +1,6 @@
 <!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
 <Project>
     <PropertyGroup>
-        <VersionPrefix>2.0.0</VersionPrefix>
+        <VersionPrefix>2.0.1</VersionPrefix>
```
bricelam: @natemcmaster What else do I need to do?

natemcmaster: Is this for the 2.0.6 patch (Feb)?

bricelam: @ajcvickers

natemcmaster: The changes in this repo are all you need. Depending on which patch release this is targeting, you also need to update aspnet/Universe and upstream repos. For example, you'll need to update aspnet/EFCore to pull this new package version.

src/Microsoft.Data.Sqlite.Core/SqliteTransaction.cs
```diff
@@ -43,9 +43,10 @@ internal SqliteTransaction(SqliteConnection connection, IsolationLevel isolation
                 throw new ArgumentException(Resources.InvalidIsolationLevel(isolationLevel));
             }
 
-            // TODO: Register transaction hooks to detect when a user manually completes a transaction created using
-            //       this API
-            connection.ExecuteNonQuery("BEGIN;");
+            connection.ExecuteNonQuery(
+                IsolationLevel == IsolationLevel.Serializable
```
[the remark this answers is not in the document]
bricelam: That's why I'm not using the parameter `isolationLevel`, but instead the property `IsolationLevel`. This will check to see what the current isolation level is.

PR #488, opened 2018-02-28
Update to SQLitePCLRaw 1.1.10-pre20180223200113

natemcmaster:
Resolves https://github.com/aspnet/EntityFrameworkCore/issues/10724

Includes support for Alpine Linux and "portable" windows RIDs (win-x64/win-x86)

cc @ericsink


natemcmaster:
From Eric's release notes, here are more changes:

> update e_sqlite3 to 3.22.0 and turn on FTS5.  
fix bundled sqlcipher build for UWP.  
AssemblyVersion now being updated properly.  
attempt fix crash involving CLR finalizer.  
add e_sqlite3 builds for linux-arm64 and alpine-x64.  
change generic Windows builds to use win-foo instead of win7-foo.  
add support for SQLITE_DETERMINISTIC.  
added sqlite3_blob_open overload to support higher perf in certain cases. 

It doesn't appear to me that any of these would affect EF Core, but would be good to run tests with EF Core just to be sure.

PR #491, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


smitpatel:
Would this be required though?
Baselines were created using 2.0.0 version. Each patch version should be non-breaking to the RTM version and will change assembly version too. Do we need to keep on changing assembly version in the file with every patch?
cc: @ajcvickers

ryanbrandenburg:
@Eilon do we want to merge these where there's no real update just to indicate that they've been checked?

Eilon:
We had said for this "refresh" we'd use the latest patch, which I think is fine. Once everything is up-to-date I don't think we'd update to any newer patches because patches shouldn't be adding new APIs (except in super rare cases).

PR #500, opened 2018-04-09
Add GetBytes, GetChars and GetStream to SqliteDataReader

AlexanderTaeschner:
Addresses parts of #18

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
 using SQLitePCL;
 
 namespace Microsoft.Data.Sqlite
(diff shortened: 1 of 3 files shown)


bricelam:
FYI, we're locked down on features for version 2.1.0, so merging will have to wait until we get a 2.2.0 branch.

src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
```diff
@@ -170,9 +171,20 @@ public static Type GetFieldType(string type)
...
         public virtual long GetChars(int ordinal, long dataOffset, char[] buffer, int bufferOffset, int length)
-            => throw new NotSupportedException();
+        {
+            var text = GetString(ordinal);
+            text.CopyTo((int)dataOffset, buffer, bufferOffset, length);
+            return length;
+        }
+
+        public virtual Stream GetStream(int ordinal)
+            => new MemoryStream(GetBlob(ordinal));
```
bricelam: Make it read-only.

bricelam: I wonder if multiple calls should share the same buffer instance.

AlexanderTaeschner: Done.

AlexanderTaeschner: Done.

PR #501, opened 2018-04-09
Add incremental blob I/O

AlexanderTaeschner:
Addresses #18

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
@@ -228,6 +228,33 @@ namespace Microsoft.Data.Sqlite
         }
 
         /// <summary>
+        /// Open a BLOB for incremental I/O.
+        /// </summary>
+        /// <param name="tableName">The table name.</param>
+        /// <param name="columnName">The column name.</param>
+        /// <param name="rowid">The row index.</param>
+        /// <param name="readWriteAccess">Flag indicating whether read and write access is desired (otherwise only read access).</param>
+        /// <returns>Object to access the BLOB.</returns>
+        public SqliteBlob OpenBlob(string tableName, string columnName, long rowid, bool readWriteAccess)
+            => OpenBlob(MainDatabaseName, tableName, columnName, rowid, readWriteAccess);
+
+        /// <summary>
+        /// Open a BLOB for incremental I/O.
+        /// </summary>
+        /// <param name="databaseName">The name of the database.</param>
+        /// <param name="tableName">The table name.</param>
+        /// <param name="columnName">The column name.</param>
+        /// <param name="rowid">The row index.</param>
+        /// <param name="readWriteAccess">Flag indicating whether read and write access is desired (otherwise only read access).</param>
+        /// <returns>Object to access the BLOB.</returns>
+        public SqliteBlob OpenBlob(string databaseName, string tableName, string columnName, long rowid, bool readWriteAccess)
+        {
+            var rc = raw.sqlite3_blob_open(_db, databaseName, tableName, columnName, rowid, readWriteAccess ? 1 : 0, out var blob);
+            SqliteException.ThrowExceptionForRC(rc, _db);
+            return new SqliteBlob(blob, _db);
+        }
+
+        /// <summary>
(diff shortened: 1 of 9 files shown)


AlexanderTaeschner:
Second commit adds call to `sqlite3_bind_zeroblob` to be able to add an empty BLOB which already has a reserved size.

src/Microsoft.Data.Sqlite.Core/SqliteConnection.cs
```diff
@@ -227,6 +227,33 @@ public override void Open()
...
 
+        /// <summary>
+        /// Open a BLOB for incremental I/O.
+        /// </summary>
+        /// <param name="tableName">The table name.</param>
+        /// <param name="columnName">The column name.</param>
+        /// <param name="rowid">The row index.</param>
+        /// <param name="readWriteAccess">Flag indicating whether read and write access is desired (otherwise only read access).</param>
+        /// <returns>Object to access the BLOB.</returns>
+        public SqliteBlob OpenBlob(string tableName, string columnName, long rowid, bool readWriteAccess)
```
bricelam: I'd rather not add new API for this.

Instead, I'd like `SqliteDataReader.GetStream()`, `GetBytes()`, and `GetChars()` to start using `sqlite3_blob` if you happen to select `rowid` (or an alias)

AlexanderTaeschner: Okay, but this means that we would not provide the possibility to add an empty blob and write to it later (or do you want to have such a functionality included in GetStream?)!

AlexanderTaeschner: At the momnet I have no idea how to recognize, that id is an alias for rowid. Any suggestions?

bricelam: Hmm... Is there anything for that in the GetSchemaTable() code? I think you can assume that if there's a single `integer` primary key column on the table, that column aliases `rowid` (unless they created it `WITHOUT ROWID` in which case they can't use blol I/o)

bricelam: There may be something in here too: https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore.Sqlite.Core/Scaffolding/Internal/SqliteDatabaseModelFactory.cs

bricelam:
I was curios how well `sqlite3_blob` mapped to `Stream`, so I wrote a bit of code in [bricelam:blob](https://github.com/aspnet/Microsoft.Data.Sqlite/compare/dev...bricelam:blob). It looks like a more direct API can also be available to do this:
``` csharp
using (var stream = new SqliteBlob(
    connection,
    "main",
    "myTable",
    "myBlobColumn",
    rowed,
    writable: true))
{
    // TODO: Use stream
}
```

AlexanderTaeschner:
Changed SqliteBlob to derive from Stream and adapted the test accordingly. The proposed changes to SqliteDataReader.GetStream(), GetBytes(), and GetChars() to start using sqlite3_blob if one happens to select rowid (or an alias) can come in a later patch.

AlexanderTaeschner:
The SqliteBlob object is now also used in GetStream. I am not sure if it makes sense to add code to GetBytes or GetChars, since the overhead for finding the rowid column (if one exists at all) is not small.

bricelam:
Sorry I haven't had a chance to dig into this yet. We're locking down on the 2.2.0 release, and I think this is one of those features we'll need a few pre-releases of to gather feedback on and find any bugs. I'll still try to merge it when I get a chance; it'll just go into `master` (the branch for 3.0.0).

ajcvickers:
@AlexanderTaeschner We are currently going through the process of reducing the number of GitHub repositories by merging some repos together. For full details see this announcement: https://github.com/aspnet/Announcements/issues/320

This means that the SQLite code will be moving to the [EntityFrameworkCore repo](https://github.com/aspnet/EntityFrameworkCore) in the next couple of days. Unfortunately, there isn't a great way for us to move open PRs over automatically, so we will be closing the PR here. We would really appreciate it if you could re-submit the PR in the new location after the move. We apologize for the hassle.

PR #509, opened 2018-04-21
Add connection configuration

AlexanderTaeschner:
Addresses #354


bricelam:
My latest thinking on this is that we should probably add a connection pool in version 3.0, and when we do that, we can make the calls to `CreateFunction()`, etc. persistent.

We're definitely not ready to tackle this item yet...

PR #510, opened 2018-04-24
[2.1] Preserve original SQL when retrying sqlite3_prepare_v2()

### LINKED ISSUE #508: Multi-threaded environment occasionally throws "bad parameter or other API misuse" Error (21)
Here I'm NOT talking about concurrent using of a single connection from many threads (which is bad), but creating and using several connections in concurrent threads. That means 1-to-1 relation between thread and connection. This should be totally legit.

If we'll be using CreateFunction in one of concurrent threads then in the others we'll get Microsoft.Data.Sqlite.SqliteException on calling `ExecuteNonQuery` (or similar methods):
SQLite Error 21: 'bad parameter or other API misuse For more information on this error code see http://sqlite.org/rescode.html'.
stack:
```
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.<PrepareAndEnumerateStatements>d__62.MoveNext()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteNonQuery()
```

Microsoft.Data.Sqlite#2.1.0-preview2.

Code that reproduces the bug we can find below.

The bug is pretty important as I can't see how to workaround it. It'd make us to add synchronization code for every command execution code. But what to do with reading via DataReader? Should we synchronize the whole reading from datareader or just its creation (ExecuteReader)? Anyway it'd pollute user code.

Demo:
```csharp
using System;
using System.Threading;
using System.IO;
using Microsoft.Data.Sqlite;

namespace sqlite
{
    class Program
    {
    	static void execute(SqliteConnection con, string cmdText) 
    	{
			var cmd = con.CreateCommand();
			cmd.CommandText = cmdText;
			cmd.ExecuteNonQuery();
    	}

    	static void registerFunctions(SqliteConnection connection)
    	{
			connection.CreateFunction("UPPER", (string str) => str?.ToUpper());
    	}

        static void Main(string[] args)
        {
        	{
        	File.Delete("test.db");
       		var con = new SqliteConnection();
     		con.ConnectionString = "Data Source=test.db";
			con.Open();
			execute(con, "create table t1(int i)");
			execute(con, "insert into t1 values (1)");
			con.Close();
			}
			for(int i = 0; i < 100; i++) {
				Console.Write(i);

				var startEvent = new ManualResetEvent(false);
				Thread t1 = new Thread(() => {
					startEvent.WaitOne();
        			var con = new SqliteConnection();
        			con.ConnectionString = "Data Source=test.db";
					con.Open();
					registerFunctions(con);
					execute(con, "insert into t1 values (" + i + " )");
					con.Close();
				});
				Thread t2 = new Thread(() => {
					startEvent.WaitOne();
	        		var con = new SqliteConnection();
        			con.ConnectionString = "Data Source=test.db";
					con.Open();
					execute(con, "insert into t1 values (" + i + " )");
					registerFunctions(con);
					con.Close();
				});
				t1.Start();
				t2.Start();
				startEvent.Set();

				t1.Join();
				t2.Join();

				Console.WriteLine("...OK");
			}

            Console.WriteLine("OK.");
        }
    }
}
```

bricelam:
Fixes #508

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
@@ -527,7 +527,8 @@ namespace Microsoft.Data.Sqlite
             var tail = _commandText;
             do
             {
-                while (IsBusy(rc = raw.sqlite3_prepare_v2(_connection.Handle, tail, out stmt, out tail)))
+                string nextTail;
+                while (IsBusy(rc = raw.sqlite3_prepare_v2(_connection.Handle, tail, out stmt, out nextTail)))
                 {
                     if (timer.ElapsedMilliseconds >= CommandTimeout * 1000)
                     {
@@ -536,6 +537,7 @@ namespace Microsoft.Data.Sqlite
 
                     Thread.Sleep(150);
                 }
+                tail = nextTail;
 
                 SqliteException.ThrowExceptionForRC(rc, _connection.Handle);
 


ajcvickers:
@bricelam Approved for RC1. Please merge this into the release/2.1 ASAP so it gets in the next prodcon build.
/cc @natemcmaster FYI.

bricelam:
P.S. I tried adding a regression test for this, but I couldn't find a way to repro it without just throwing a bunch of threads at it and waiting for it to happen.

I verified that the original repro is fixed by this change.

PR #515, opened 2018-05-06
Simple overflow checking on GetBytes and GetChars

mlohstroh:
- With the methods that return the number of bytes that are read, it should be expected that they should safeguard against reading over the buffer length especially since the user might not know the length beforehand. 
- Before this patch, if you stuck in a length that was larger than the rest of the buffer size, the `Array.Copy` would throw an argument exception. Now, after this patch it behaves like `Stream.Read`.

There didn't seem to be an bug# associated with this. I found it when integrating the latest nightly into my app.


bricelam:
Thanks!

PR #519, opened 2018-05-16
Upgrade to netcoreapp22

ryanbrandenburg:


ryanbrandenburg:
There's an issue filed for that: aspnet/Universe#1153. We're trying to minimize the work involved here since it's already taken a long time.

PR #533, opened 2018-06-27
Handle externally rolled back transactions

### LINKED ISSUE #532: Transactions don't handle RAISE(ROLLBACK) in triggers
The following code throws *cannot rollback - no transaction is active*.

``` csharp
var connection = new SqliteConnection("Filename=:memory:");
connection.Open();

command.CommandText =
@"
    CREATE TABLE test(value);

    CREATE TRIGGER test_trigger
    BEFORE INSERT ON test
    BEGIN
        SELECT RAISE(ROLLBACK, 'error');
    END;
";
command.ExecuteNonQuery();

var transaction = connection.BeginTransaction();
try
{
    command.Transaction = transaction;
    command.CommandText = "INSERT INTO test VALUES ('data')";
    command.ExecuteNonQuery();
}
catch (SqliteException ex)
    when (ex.SqliteErrorCode == SQLitePCL.raw.SQLITE_CONSTRAINT)
{
}
finally
{
    transaction.Rollback(); // cannot rollback - no transaction is active
}
```

We should leverage `sqlite3_rollback_hook` to detect this and similar situations and try to handle them more gracefully.

bricelam:
Fixes #532

### DIFF
--- a/test/Microsoft.Data.Sqlite.Tests/SqliteTransactionTest.cs
+++ b/test/Microsoft.Data.Sqlite.Tests/SqliteTransactionTest.cs
@@ -178,6 +178,24 @@ namespace Microsoft.Data.Sqlite
         }
 
         [Fact]
+        public void Commit_throws_when_completed_externally()
+        {
+            using (var connection = new SqliteConnection("Data Source=:memory:"))
+            {
+                connection.Open();
+
+                using (var transaction = connection.BeginTransaction())
+                {
+                    connection.ExecuteNonQuery("ROLLBACK;");
+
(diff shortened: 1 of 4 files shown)


test/Microsoft.Data.Sqlite.Tests/SqliteTransactionTest.cs
```diff
@@ -177,6 +177,24 @@ public void Commit_throws_when_completed()
             }
         }
 
+        [Fact]
+        public void Rollback_throws_when_completed_externally()
```
bricelam: 😄 Should be `Commit_throws`...

Issue #422, opened 2017-09-07
Wrong number of rows returned when batching operations


Grauenwolf:
If I execute this command I get 5 rows back.

```
SELECT [EmployeeKey], [FirstName], [MiddleName], [LastName], [Title], [ManagerKey], [CreatedDate], [UpdatedDate] FROM Employee WHERE EmployeeKey IN (@Param0, @Param1, @Param2, @Param3, @Param4);
@Param0: 11
@Param1: 12
@Param2: 13
@Param3: 14
@Param4: 15
```

If I batch it with another command, I only get 1 row back. (EmployeeKey 11 in this case.)

```
SELECT [EmployeeKey], [FirstName], [MiddleName], [LastName], [Title], [ManagerKey], [CreatedDate], [UpdatedDate] FROM Employee WHERE EmployeeKey IN (@Param0, @Param1, @Param2, @Param3, @Param4);
DELETE FROM Employee WHERE EmployeeKey IN (@Param0, @Param1, @Param2, @Param3, @Param4);
@Param0: 11
@Param1: 12
@Param2: 13
@Param3: 14
@Param4: 15
```

This only happens with Microsoft.Data.Sqlite. If I run the same batch with System.Data.SQLite I get back the expected number of rows.

Grauenwolf:
Here's my package list in case its relevant.

```
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="Microsoft.Bcl" version="1.1.10" targetFramework="net461" />
  <package id="Microsoft.Bcl.Async" version="1.0.168" targetFramework="net461" />
  <package id="Microsoft.Bcl.Build" version="1.0.21" targetFramework="net461" />
  <package id="Microsoft.Data.Sqlite" version="2.0.0" targetFramework="net461" />
  <package id="Microsoft.Data.Sqlite.Core" version="2.0.0" targetFramework="net461" />
  <package id="Microsoft.NETCore.Platforms" version="1.0.1" targetFramework="net461" />
  <package id="Microsoft.Win32.Primitives" version="4.0.1" targetFramework="net461" />
  <package id="MSTest.TestAdapter" version="1.1.18" targetFramework="net461" />
  <package id="MSTest.TestFramework" version="1.1.18" targetFramework="net461" />
  <package id="Nito.AsyncEx" version="4.0.1" targetFramework="net461" />
[...]

PR #541, opened 2018-07-26
Fix and Modify some codes to make it better and more strictly

Maledong:
1) Use IsNullOrWhiteSpace instead of IsNullOrEmpty for SqliteParameter's
name and SqlCommand's _commandText, because we don't wanna see names full of white spaces or invisible symbols (They are meaningless at all).

2) Use `using.....` instead of exclipitly calling `Dispose()`.

3) Add a space in the comment for `BulkInsertSample`.

### DIFF
--- a/samples/BulkInsertSample/Program.cs
+++ b/samples/BulkInsertSample/Program.cs
@@ -42,7 +42,7 @@ namespace BulkInsertSample
                 insertCommand.Parameters.Add(valueParameter);
 
                 // No need to call Prepare() since it's done lazily during the first execution.
-                //insertCommand.Prepare();
+                // insertCommand.Prepare();
 
                 var random = new Random();
 
--- a/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
@@ -230,7 +230,7 @@ namespace Microsoft.Data.Sqlite
                 throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(Prepare)));
             }
 
-            if (string.IsNullOrEmpty(_commandText))
+            if (string.IsNullOrWhiteSpace(_commandText))
             {
                 throw new InvalidOperationException(Resources.CallRequiresSetCommandText(nameof(Prepare)));
             }
@@ -298,7 +298,7 @@ namespace Microsoft.Data.Sqlite
                 throw new InvalidOperationException(Resources.CallRequiresOpenConnection(nameof(ExecuteReader)));
             }
 
-            if (string.IsNullOrEmpty(_commandText))
+            if (string.IsNullOrWhiteSpace(_commandText))
             {
                 throw new InvalidOperationException(Resources.CallRequiresSetCommandText(nameof(ExecuteReader)));
             }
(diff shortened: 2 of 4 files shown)


src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -246,9 +246,7 @@ public override void Prepare()
             {
                 using (var enumerator = PrepareAndEnumerateStatements(timer).GetEnumerator())
                 {
-                    while (enumerator.MoveNext())
```
smitpatel: We don't use this coding pattern. Please review coding guidelines of this project first.

Maledong: Fixed it, thanks!

Maledong:
@smitpatel：Thanks!
BTW：There's something wrong with Travis-ci when generating scripts for Mac, so would you mind restarting this check again? It went well with me yesterday.

Maledong:
@smitpatel：OK, fixed. Thanks!

bricelam:
I take personal issue with `IsNullOrWhiteSpace`. I don't like that it adds additional ops for every valid case just to catch a few more obscure invalid cases.

bricelam:
Did you encounter an error when using whitespace in these codepaths?

samples/BulkInsertSample/Program.cs
```diff
@@ -42,7 +42,7 @@ value INTEGER
                 insertCommand.Parameters.Add(valueParameter);
 
                 // No need to call Prepare() since it's done lazily during the first execution.
-                //insertCommand.Prepare();
+                // insertCommand.Prepare();
```
bricelam: This has different semantics to me. Before it was commented out code, now it's a comment showing a code example--an example of something you shouldn't do, no less.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
```diff
@@ -479,15 +479,15 @@ public override int ExecuteNonQuery()
...
             }
 
-            var reader = ExecuteReader();
-            reader.Dispose();
-
-            return reader.RecordsAffected;
+            using (var reader = ExecuteReader())
+            {
+                return reader.RecordsAffected;
+            }
```
bricelam: Hmm, this might result in a different value. `RecordsAffected` may not include all the results until all the result sets are consumed.

bricelam: (It certainly will after we fix #422)

samples/BulkInsertSample/Program.cs
Maledong: Sorry but I don't mean this is an example but just to make the alignment of vertical, which makes codes look beautiful.

src/Microsoft.Data.Sqlite.Core/SqliteCommand.cs
Maledong: Well, this just make your codes in union like what you see line at 509~515, which uses `using....` statement to wrap those implementing IDisposible.

Maledong:
@bricelam：The biggest problem that I think is sometimes the codes aren't so in union.

If you take a close look at what I modified in the history, you'll find that some connections cannot be null, empty or full of white spaces. And sometimes it refers `IsNullOrEmpty`, while the others `if (_connectionString==null)`. On the other hand, since we've got IsNullOrWhiteSpace(), why don't we directly use that to check the value, this is an easy thing by hand.

So I just make your codes in union with the same condition check, this will be easy to look up and do checking for your next time.

PR #545, opened 2018-09-06
Check for null in buffer[] and return appropriate value

jpfreyen:
I'd like to submit this as fix to the 2.2.0 release.

Have GetBytes() handle buffer null value appropriately without the need for code external to this API to do so.

Just in case I messed up the PR, here is the code change:

diff --git a/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs b/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
index 9748cb1..b7d674b 100644
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
@@ -178,7 +178,8 @@ public virtual long GetBytes(int ordinal, long dataOffset, byte[] buffer, int bu

             long bytesToRead = (long)blob.Length - dataOffset;
             bytesToRead = System.Math.Min(bytesToRead, length);
-            Array.Copy(blob, dataOffset, buffer, bufferOffset, bytesToRead);
+            if (buffer != null)
+                Array.Copy(blob, dataOffset, buffer, bufferOffset, bytesToRead);
             return bytesToRead;


jpfreyen:
This isn't right, it should only be a single file with a single patch...I was hoping I could just attach the patch.  I'll cancel and try again.

PR #546, opened 2018-09-06
Check for null in buffer[] and return appropriate value

jpfreyen:
OK, this looks better.

Have GetBytes() handle buffer null value appropriately without the need for code external to this API to do so.

I'd like to get this fix into 2.2.0 release.  Thank you.

### DIFF
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
@@ -178,7 +178,8 @@ namespace Microsoft.Data.Sqlite
 
             long bytesToRead = (long)blob.Length - dataOffset;
             bytesToRead = System.Math.Min(bytesToRead, length);
-            Array.Copy(blob, dataOffset, buffer, bufferOffset, bytesToRead);
+            if (buffer != null)
+                Array.Copy(blob, dataOffset, buffer, bufferOffset, bytesToRead);
             return bytesToRead;
         }
 
--- a/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
+++ b/src/Microsoft.Data.Sqlite.Core/SqliteDataRecord.cs
@@ -177,9 +177,11 @@ namespace Microsoft.Data.Sqlite
             var blob = GetCachedBlob(ordinal);
 
             long bytesToRead = (long)blob.Length - dataOffset;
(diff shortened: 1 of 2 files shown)


smitpatel:
@jpfreyen - Can you add regression test? Also is there existing bug tracking this issue?

jpfreyen:
I could add a test, but I don't know how and I'm not familiar enough with the project to add one.

Same with bug tracking...I didn't file a bug because I don't know the bug tracking mechanism to use.

smitpatel:
Looking at this commit will give you more info.
https://github.com/aspnet/Microsoft.Data.Sqlite/commit/f975baa9a8d91ee91dc55e7cd7c4b0ee431f9aa5
It is the same commit which added the line where you need null check as in this PR.
You should be able to test similar to that in SqliteDataReaderTest.cs which would have null buffer and needs above check.

jpfreyen:
I filed a ticket #547 and pointed it to this issue; but I'm showstopped to write a unit test.  Hope that's not a blocker to get this fix in.

smitpatel:
VS for Mac likely not work properly. All ASP.NET repos uses preview build of .NET core SDK. Currently VS for Mac is not good at recognizing preview SDK and using them whenever asked. In VS for windows, it works because VS picks up SDK based on ENV:Path variable.

Running `build.sh` in repo root would build whole project and run all tests. You can also add preview SDK to your path. (build script installs it so you know where its located), and then call `dotnet test` on test project's csproj file directly to run test on that project only. This article describes how to run selective tests using `dotnet test`. https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests

jpfreyen:
K, I added a test.

I also made a tweak to GetBytes() null handling, based on
https://www.devart.com/dotconnect/sqlite/docs/Devart.Data.SQLite~Devart.Data.SQLite.SQLiteDataReader~GetBytes(Int32,Int64,Byte[],Int32,Int32).html

"GetBytes returns the number of available bytes in the field. In most cases this is the exact length of the field. However, the number returned may be less than the true length of the field if GetBytes has already been used to obtain bytes from the field. This may be the case, for example, if the SQLiteDataReader is reading a large data structure into a buffer.
**If you pass a buffer that is a null value, GetBytes returns the length of the field in bytes**."

bricelam:
@jpfreyen Thanks for the contribution! Sorry it took me so long to merge it.

jpfreyen:
no problem :-)

Issue #549, opened 2018-09-07
GetCachedBlob() can cause repeat of data to be read incorrectly


jpfreyen:
This was found in the `origin/release/2.2` tag/release.

Let's see if I can explain this right, as it is odd behavior.

In general, if I have a SQL statement that returns more than one row, and the pseudocode below is used to read a blob from each row, 

```
while "DataRecord Object".Read() {
   "DataRecord Object".GetBytes(ordinal, dataOffset, buffer, bufferOffset, length);
}
```

you will get back the same data blob each iteration.  This is because the blob is cached within the SqliteDataRecord.cs in GetBytes():

`  var blob = GetCachedBlob(ordinal);
`

since `_blobCache[ordinal];` is not cleared for each SQL Data Reader Read (.Read()). (i.e., it always goes to the cache pulling out the old value instead of the next value which is within the underlying `_stmt`).

A suggested fix is of follows:

1. In SqliteDataRecord.cs define a new method:

```
public void ClearBlobCache() {
   _blobCache = new byte[FieldCount][];
}
```
2. In the SqliteDataRecord() constructor, add:

`ClearBlobCache();
`
and remove the line:

`_blobCache = new byte[FieldCount][];
`

3.  In SqliteDataReader.cs in Read(), before the "return !_done;" statement, add to make sure that the data is reset:

```
if !_done
   _record.ClearBlobCache();
```

I hope this gives the general gist what the problem is, even if this isn't the appropriate fix for the issue.

PR #550, opened 2018-09-08
Clear cached data records while stepping.

AlexanderTaeschner:
Addresses #549.


bricelam:
(Rebased on top of release/2.2)

PR #551, opened 2018-09-21
Remove un-needed framework references

ryanbrandenburg:
@bricelam for best guess on who might "own" this repo. If it's not you please let me know who would be better suited.

We're removing these extra framework referenced as part of [investigating this](https://github.com/aspnet/Coherence-Signed/issues/1062#issuecomment-423590031) to see if it resolves the issue.


bricelam:
Seems fine. So long as we keep testing on netcoreapp3.0 for now (to catch regressions), I'm happy with any tweaks.

bricelam:
Interesting. Oh well. Will sort out when we start focusing on 3.0

